PR #30, opened 2015-04-29
Replace intval() with faster (int) cast

thiemowmde:
intval() is about 5 times slower than a type cast, no matter if it's done explicit (by explicitly stating (int)$string) or implicit (via $string > 0). The difference between the two is that intval() accepts a base, which is not needed in these cases.

Following the arguments given in https://forge.typo3.org/issues/54265.

Additional changes:
* Use non-fetching `(?:)` when that fragment is not needed.
* Avoid calling the way to complicated (and slow) normalize method when the only relevant character is the comma.
* Make stuff private by default.
* Optimize imports.

### DIFF
--- a/src/ValueParsers/DecimalParser.php
+++ b/src/ValueParsers/DecimalParser.php
@@ -5,7 +5,6 @@ namespace ValueParsers;
 use DataValues\DecimalMath;
 use DataValues\DecimalValue;
 use DataValues\IllegalValueException;
-use InvalidArgumentException;
 
 /**
  * ValueParser that parses the string representation of a decimal number.
@@ -27,7 +26,7 @@ class DecimalParser extends StringValueParser {
 	/**
 	 * @var null|NumberUnlocalizer
 	 */
-	protected $unlocalizer;
+	private $unlocalizer;
 
 	/**
 	 * @since 0.1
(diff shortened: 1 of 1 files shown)


src/ValueParsers/DecimalParser.php
```diff
@@ -66,9 +65,9 @@ private function getMath() {
 	 *         does not use scientific notation, $exponent will be 0.
 	 */
 	public function splitDecimalExponent( $valueString ) {
-		if ( preg_match( '/^(.*)([eE]|x10\^)([-+]?[,\d]+)$/', $valueString, $matches ) ) {
-			$exponent = $this->normalizeDecimal( $matches[3] );
-			return array( $matches[1], intval( $exponent ) );
+		if ( preg_match( '/^(.*)(?:[eE]|x10\^)([-+]?[\d,]+)$/', $valueString, $matches ) ) {
```
mariushoch: I wonder why we allow commas here... there aren't any tests for this method, so I have no idea what kind of input would that be. Also Wikibase's MediaWikiNumberUnlocalizer::getNumberRegex doesn't match commas in the exponent, while it says "Keep in sync with DecimalParser::splitDecimalExponent.". Unrelated to this change, but I wanted to let you know.

thiemowmde: The only reason I can think of is that @brightbyte wanted to make this parser a little bit more flexible so it accepts exponents with group separators. This is tested, see `12e3,4`.

brightbyte: Yes, that was the reason. 2.3E27,428 may look odd, but it's valid, as far as I know.

brightbyte:
Is there a specific reason for this micro-optimization, btw? I seems a bit pointless to me.

thiemowmde:
Which one?

PR #31, opened 2015-06-08
Introduce QuantityUnitFormatter.

brightbyte:
This introduces an interface for formatting quantity units.

NOTE: This is a BREAKING CHANGE, since it adds a requried
parameter to the constructor of QuantityFormatter.

[Bug: T92380](https://phabricator.wikimedia.org/T92380)

### DIFF
--- a/src/ValueFormatters/BasicQuantityUnitFormatter.php
+++ b/src/ValueFormatters/BasicQuantityUnitFormatter.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace ValueFormatters;
+
+/**
+ * A basic QuantityUnitFormatter.
+ *
+ * @licence GNU GPL v2+
+ * @author Daniel Kinzler
+ */
+class BasicQuantityUnitFormatter implements QuantityUnitFormatter {
+
+	/**
+	 * @see QuantityUnitFormatter::applyUnit
+	 *
+	 * This basic implementation simply appends $unit to $numberText,
+	 * unless $unit is "1" or "", in which case  $numberText is returned unmodified.
+	 *
+	 * @param string $unit
+	 * @param string $numberText
+	 *
+	 * @return string A text representing the given number with the given unit applied.
+	 */
+	public function applyUnit( $unit, $numberText ) {
+		if ( $unit === '1' || $unit === '' ) {
+			return $numberText;
+		} else {
+			return $numberText . $unit;
+		}
+	}
+}
--- a/src/ValueFormatters/BasicQuantityUnitFormatter.php
+++ b/src/ValueFormatters/BasicQuantityUnitFormatter.php
@@ -28,4 +28,5 @@ class BasicQuantityUnitFormatter implements QuantityUnitFormatter {
 			return $numberText . $unit;
 		}
 	}
+
 }
--- a/src/ValueFormatters/QuantityFormatter.php
+++ b/src/ValueFormatters/QuantityFormatter.php
@@ -38,9 +38,20 @@ class QuantityFormatter extends ValueFormatterBase {
 	const OPT_APPLY_ROUNDING = 'applyRounding';
 
 	/**
+	 * Option key controlling whether the quantity's unit of measurement should be included
+	 * in the output.
+	 */
+	const OPT_APPLY_UNIT = 'applyUnit';
+
+	/**
 	 * @var DecimalMath
 	 */
 	protected $decimalMath;
+
+	/**
+	 * @var QuantityUnitFormatter
+	 */
+	private $unitFormatter;
 
 	/**
 	 * @var DecimalFormatter
@@ -50,15 +61,17 @@ class QuantityFormatter extends ValueFormatterBase {
 	/**
 	 * @param DecimalFormatter|null $decimalFormatter
 	 * @param FormatterOptions|null $options
+	 * @param QuantityUnitFormatter $unitFormatter
 	 */
-	// FIXME: In all other ValueFormatters the FormatterOption parameter is the first one.
-	public function __construct( DecimalFormatter $decimalFormatter = null, FormatterOptions $options = null ) {
+	public function __construct( QuantityUnitFormatter $unitFormatter, DecimalFormatter $decimalFormatter = null, FormatterOptions $options = null ) {
 		parent::__construct( $options );
 
 		$this->defaultOption( self::OPT_SHOW_UNCERTAINTY_MARGIN, true );
 		$this->defaultOption( self::OPT_APPLY_ROUNDING, true );
+		$this->defaultOption( self::OPT_APPLY_UNIT, true );
 
 		$this->decimalFormatter = $decimalFormatter ?: new DecimalFormatter( $this->options );
+		$this->unitFormatter = $unitFormatter;
 
 		// plain composition should be sufficient
 		$this->decimalMath = new DecimalMath();
@@ -120,16 +133,15 @@ class QuantityFormatter extends ValueFormatterBase {
 			}
 		}
 
-		//TODO: use localizable pattern for constructing the output.
 		$quantity = $amount;
 
 		if ( $margin !== '' ) {
+			//TODO: use localizable pattern for constructing the output.
 			$quantity .= '±' . $margin;
 		}
 
-		if ( $unit !== '1' ) {
-			//XXX: do we need to localize unit names?
-			$quantity .= $unit;
+		if ( $this->options->getOption( self::OPT_APPLY_UNIT ) && $unit !== '1' && $unit !== '' ) {
+			$quantity = $this->unitFormatter->applyUnit( $unit, $quantity );
 		}
 
 		return $quantity;
--- a/src/ValueFormatters/QuantityFormatter.php
+++ b/src/ValueFormatters/QuantityFormatter.php
@@ -49,21 +49,25 @@ class QuantityFormatter extends ValueFormatterBase {
 	protected $decimalMath;
 
 	/**
-	 * @var QuantityUnitFormatter
-	 */
-	private $unitFormatter;
-
-	/**
 	 * @var DecimalFormatter
 	 */
 	protected $decimalFormatter;
 
 	/**
-	 * @param DecimalFormatter|null $decimalFormatter
-	 * @param FormatterOptions|null $options
-	 * @param QuantityUnitFormatter $unitFormatter
+	 * @var QuantityUnitFormatter
 	 */
-	public function __construct( QuantityUnitFormatter $unitFormatter, DecimalFormatter $decimalFormatter = null, FormatterOptions $options = null ) {
+	private $unitFormatter;
+
+	/**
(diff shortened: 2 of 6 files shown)


src/ValueFormatters/BasicQuantityUnitFormatter.php
```diff
@@ -0,0 +1,33 @@
...
+	 * @param string $numberText
+	 *
+	 * @return string A text representing the given number with the given unit applied.
+	 */
+	public function applyUnit( $unit, $numberText ) {
+		if ( $unit === '1' || $unit === '' ) {
+			return $numberText;
+		} else {
+			return $numberText . $unit;
+		}
```
JeroenDeDauw: No if-else is needed here. A simple guard will do

JeroenDeDauw:
Looks sane design wise

src/ValueFormatters/BasicQuantityUnitFormatter.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace ValueFormatters;
+
+/**
+ * Basic (dummy) unit formatter.
```
thiemowmde: Why "dummy"? It's not that this implementation doesn't do anything. It just misses localization.

src/ValueFormatters/BasicQuantityUnitFormatter.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @see ValueParsers\Unlocalizer
```
thiemowmde: Copy-pasto. this doesn't even exist any more.

src/ValueFormatters/BasicQuantityUnitFormatter.php
```diff
@@ -0,0 +1,33 @@
...
+	 * @param string $unit
+	 * @param string $numberText
+	 *
+	 * @return string A text representing the given number with the given unit applied.
+	 */
+	public function applyUnit( $unit, $numberText ) {
+		if ( $unit === '1' || $unit === '' ) {
+			return $numberText;
+		} else {
+			return $numberText . $unit;
```
thiemowmde: Why not make the two distinguishable with a space?

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -50,15 +56,21 @@ class QuantityFormatter extends ValueFormatterBase {
 	/**
 	 * @param DecimalFormatter|null $decimalFormatter
 	 * @param FormatterOptions|null $options
+	private $unitFormatter;
```
thiemowmde: Rebase mistake.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -50,15 +56,21 @@ class QuantityFormatter extends ValueFormatterBase {
...
+	 * @param QuantityUnitFormatter $unitFormatter
 	 */
-	// FIXME: In all other ValueFormatters the FormatterOption parameter is the first one.
 	public function __construct( DecimalFormatter $decimalFormatter = null, FormatterOptions $options = null ) {
+		$options->defaultOption( self::OPT_APPLY_UNIT, true );
 		parent::__construct( $options );
 
 		$this->defaultOption( self::OPT_SHOW_UNCERTAINTY_MARGIN, true );
 		$this->defaultOption( self::OPT_APPLY_ROUNDING, true );
+		$options->defaultOption( self::OPT_APPLY_UNIT, true );
```
thiemowmde: Same default option set two times.

tests/ValueFormatters/QuantityFormatterTest.php
```diff
@@ -67,6 +71,10 @@ public function validProvider() {
 			DecimalFormatter::OPT_FORCE_SIGN => true,
 		) );
 
+		$noUnit= new FormatterOptions( array(
```
thiemowmde: Space.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -38,6 +38,12 @@ class QuantityFormatter extends ValueFormatterBase {
 	const OPT_APPLY_ROUNDING = 'applyRounding';
 
 	/**
+	 * Option key controlling whether the quantity's unit of measurement should be included
+	 * in the output.
+	 */
+	const OPT_APPLY_UNIT = 'applyUnit';
```
thiemowmde: Why does this need to be an option? In my opinion a quantity formatter must always append the unit.

src/ValueFormatters/BasicQuantityUnitFormatter.php
brightbyte: I like it nice any symmetric like this.

src/ValueFormatters/QuantityFormatter.php
brightbyte: Not when generating a value for use in an input field. We cannot parse localized units (yet). So in case the unit is to be input separately, it must not be present in the text representation of the quantity.

tests/ValueFormatters/QuantityFormatterTest.php
brightbyte: The final frontier...

src/ValueFormatters/QuantityFormatter.php
thiemowmde: This should use a DecimalFormatter then.

brightbyte: Can't: DecimalFormat cannot format a QuantityValue. And DecimalFormat would lose the uncertainty interval (which we can and do parse).

src/ValueFormatters/QuantityFormatter.php
brightbyte: Removed the other instance.

src/ValueFormatters/QuantityFormatter.php
thiemowmde: Hm, ok. Note that one reason for my confusion is the complexity of `QuantityFormatter::format`. This method should be simple.

src/ValueFormatters/BasicQuantityUnitFormatter.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace ValueFormatters;
+
+/**
+ * A basic QuantityUnitFormatter.
```
thiemowmde: Now this line adds zero information.

src/ValueFormatters/QuantityUnitFormatter.php
```diff
@@ -0,0 +1,26 @@
...
+	 *
+	 * The unit may be placed before or after the number, with or without a separator.
+	 * The unit may also be omitted in case the $unit represents "no unit".
+	 *
+	 * @param string $unit
+	 * @param string $numberText
+	 *
+	 * @return string A text representing the given number with the given unit applied.
+	 */
+	public function applyUnit( $unit, $numberText );
```
thiemowmde: It's called "amount" and "unit" in `QuantityValue`. Having a "number text" is a bit confusing. What about `$formattedAmount` or something like that?

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -50,15 +61,17 @@ class QuantityFormatter extends ValueFormatterBase {
...
-	public function __construct( DecimalFormatter $decimalFormatter = null, FormatterOptions $options = null ) {
+	public function __construct( QuantityUnitFormatter $unitFormatter, DecimalFormatter $decimalFormatter = null, FormatterOptions $options = null ) {
 		parent::__construct( $options );
 
 		$this->defaultOption( self::OPT_SHOW_UNCERTAINTY_MARGIN, true );
 		$this->defaultOption( self::OPT_APPLY_ROUNDING, true );
+		$options->defaultOption( self::OPT_APPLY_UNIT, true );
 
 		$this->decimalFormatter = $decimalFormatter ?: new DecimalFormatter( $this->options );
+		$this->unitFormatter = $unitFormatter;
```
thiemowmde: You can avoid a breaking change with `$this->unitFormatter = $unitFormatter ?: new BasicQuantityUnitFormatter();`

brightbyte: Only if I add the new parameter at the end. Not sure that's a good idea. The thought of new extension code ending up using old library code is a bit scary.

src/ValueFormatters/QuantityUnitFormatter.php
brightbyte: $number text is *not* the amount. It's the result of formatting the amount and the uncertainty margin, so it's something like this (in German locale): 12,7+/-0,3

src/ValueFormatters/QuantityFormatter.php
brightbyte: Why should it be simple? A Quantity is a complex object consisting of a three decimal values and a unit. DecimalFormatter::format is simple.

src/ValueFormatters/QuantityFormatter.php
thiemowmde: Uh, $options can be null. :-(

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -50,15 +61,17 @@ class QuantityFormatter extends ValueFormatterBase {
 	/**
 	 * @param DecimalFormatter|null $decimalFormatter
 	 * @param FormatterOptions|null $options
+	 * @param QuantityUnitFormatter $unitFormatter
 	 */
-	// FIXME: In all other ValueFormatters the FormatterOption parameter is the first one.
-	public function __construct( DecimalFormatter $decimalFormatter = null, FormatterOptions $options = null ) {
+	public function __construct( QuantityUnitFormatter $unitFormatter, DecimalFormatter $decimalFormatter = null, FormatterOptions $options = null ) {
```
thiemowmde: The order of things confuses me. Why is the unit first? But last down in the constructor code?

src/ValueFormatters/QuantityFormatter.php
thiemowmde: The method can be split a lot. I will try to come up with a good suggestion in an other patch.

src/ValueFormatters/QuantityUnitFormatter.php
thiemowmde: I just find the name odd. A number is a number. Text is text. Moreover, this code does not care what the string contains. Which makes me think why the string is passed to this class in the first place. I think the responsibility of concatenating the two strings should not be here but in the QuantityFormatter. So I suggest to change this class into a UnitFormatter with a format method that gets a string and returns a string.

thiemowmde:
@brightbyte I squashed your three commits and added a minor one on top. Please have a look and tell me if it's ok. I will merge then.

thiemowmde:
+2 when tests pass and @brightbyte is ok with my additions. Please also review #34.

thiemowmde:
Scrutinizer is not crucial, just ignore it.

PR #33, opened 2015-06-08
Simplify regular expressions

thiemowmde:
This simplifies all regular expressions in this component, partly to increase performance (e.g. do not match parts that are not used afterwards), partly for readability (e.g. use the `\d` shortcut).


thiemowmde:
Rebased.

PR #35, opened 2015-06-11
Release 0.5.0

thiemowmde:
Please have a look at https://github.com/DataValues/Number/compare/0.4.1...master and check that I did not forget anything.

~~Please note that I think the actual 0.5.0 release is blocked because of [T95425](https://phabricator.wikimedia.org/T95425).~~ Ok, the issue already existed in 0.4 (see https://github.com/DataValues/Number/blob/0.4.1/tests/ValueFormatters/QuantityFormatterTest.php#L62), so this can't be a reason to block 0.5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,24 @@ DataValues Number has been written by Daniel Kinzler, as [Wikimedia Germany]
 
 ## Release notes
 
+### 0.5.0 (2015-06-11)
+
+#### Breaking changes
+* `QuantityFormatter` constructor parameters changed in an incompatible way
+
+#### Additions
+* Added `QuantityUnitFormatter` interface
+* Added `BasicQuantityUnitFormatter`
+* Added `QuantityFormatter::OPT_APPLY_UNIT` option
+* Added `QuantityParser::OPT_UNIT` option
+* Added `DecimalParser::applyDecimalExponent`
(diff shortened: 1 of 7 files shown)


thiemowmde:
@brightbyte, @JanZerebecki, want to do the release? I think it's ready.

README.md
```diff
@@ -50,6 +50,24 @@ DataValues Number has been written by Daniel Kinzler, as [Wikimedia Germany]
...
+* Added `BasicQuantityUnitFormatter`
+* Added `QuantityFormatter::OPT_APPLY_UNIT` option
+* Added `QuantityParser::OPT_UNIT` option
+* Added `DecimalParser::applyDecimalExponent`
+* Added `DecimalParser::splitDecimalExponent`
+
+#### Other changes
+* `QuantityParser` now correctly detect precision for scientific notation
+* Made constructor parameters optional in `DecimalFormatter` and `QuantityFormatter`
+* Updated DataValues Interfaces dependency to 0.1.5
```
JeroenDeDauw: This is not really relevant to users. This change does not require action from their side nor should it cause any sort of problems

thiemowmde: Right, it *should* not matter, but I still think it's good to know.

PR #37, opened 2015-08-13
Allow URIs as units in QuantityParser

thiemowmde:
Pinging @Jonas-WMDE as well as @brightbyte.

[Bug: T98172](https://phabricator.wikimedia.org/T98172)

### DIFF
--- a/src/ValueParsers/QuantityParser.php
+++ b/src/ValueParsers/QuantityParser.php
@@ -158,7 +158,7 @@ class QuantityParser extends StringValueParser {
 				. '|(?:\+/?-|±)\s*(' . $numberPattern . ')' // $3: plus/minus offset (uncertainty margin)
 				. '|' // or nothing
 			. ')'
-			. '\s*(' . $unitPattern . ')?' // $4: unit
+			. '\s*(\bhttps?://\S+|' . $unitPattern . ')?' // $4: unit
 			. '\s*$@u';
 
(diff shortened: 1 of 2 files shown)


brightbyte:
I'm not convinced we actually want to allow URIs to represent units in "human" input. What's the use case?

thiemowmde:
What? Sorry, I don't get that. Units **are** URIs. Why does this parser **actively blocks** entering units? What could be more specific than the unit's URI itself? This is just a bug, nothing you could disagree with. How could you disagree with the fact that a user wants to enter a unit?

brightbyte:
URIs are one way to identify units internally (Wikibase commits to using URIs, but QuantityValue doesn't care). I do not see why it should be a requirement for a parser to allow users to enter a unit using the *internal* identifier. 

We *could* allow URIs to be entered directly. I'm not totally opposed to that. But I'm not convinced that we should, either. I don't see the use case in practice.

thiemowmde:
How is this relevant for this bugfix? The parser is broken. I'm fixing it. What's the problem? How can one disagree with a user wanting to enter a unit? To what does "internal" refer to? Since when are entity ids "internal"? I remember you proposing to make them external just everywhere.

thiemowmde:
Oh, and why on earth should a **parser** actively block something that's supported by the datavalue and even **required** by the validator? That's a bug. Reminds me of the arguments we have about https://gerrit.wikimedia.org/r/219016, but with switched roles.

PR #38, opened 2015-08-14
Release 0.6.0

thiemowmde:
Please merge ~~#37 and~~ this and release 0.6.0. The release is required for https://gerrit.wikimedia.org/r/235495.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,18 +50,22 @@ DataValues Number has been written by Daniel Kinzler, as [Wikimedia Germany]
 
 ## Release notes
 
-### 0.6.0 (2015-09-04)
+### 0.6.0 (2015-09-09)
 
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -22,7 +22,7 @@
 	"require": {
 		"php": ">=5.3.0",
 		"data-values/data-values": "~1.0|~0.1",
-		"data-values/interfaces": "^0.1.5",
+		"data-values/interfaces": "~0.2.0|^0.1.5",
```
JeroenDeDauw: This was wrong before the change. Should be `~0.1.5`. Else it's `>=0.1.5 AND <1.0.0`, which happens to include the added range here, which thus has no effect

PR #40, opened 2015-08-27
MatrixValue data type

merkys:
I would like to propose addition of ``MatrixValue`` data type, that could be used to present transformation matrices, symmetry operators, tensors etc.

### DIFF
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -1,0 +1,53 @@
+<?php
+namespace DataValues;
+/**
+ * Class representing a matrix.
+ *
+ * @since 0.1
+ *
+ * @licence GNU GPL v2+
+ * @author Andrius Merkys < andrius.merkys@gmail.com >
+ */
+class MatrixValue extends DataValueObject {
+	/**
+	 * @var int|float
+	 */
+	private $value;
+
+	/**
+	 * @var int
+	 */
+	private $rows;
+
+	/**
+	 * @var int
+	 */
+	private $columns;
+
+	/**
+	 * Constructs a new MatrixValue object, representing the given value.
+	 *
+	 * @var NumberValue $value An array of numbers.
+	 * @var int $rows The number of rows.
+	 * @var int $columns The number of columns.
+	 *
+	 * @throws IllegalValueException
+	 */
+	public function __construct( NumberValue $value, $rows, $columns ) {
+		if ( $rows <= 0 ) {
+			throw new IllegalValueException( '$rows must be > 0' );
+		}
+
+		if ( $columns <= 0 ) {
+			throw new IllegalValueException( '$columns must be > 0' );
+		}
+
+        if ( count( $value ) != $rows * $columns ) {
+            throw new IllegalValueException( '$value must be of length $rows * $columns' );
+        }
+
+		$this->value = $value;
+		$this->rows = $rows;
+		$this->columns = $columns;
+	}
+}
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -64,6 +64,10 @@ class MatrixValue extends DataValueObject {
             $value_now[] = $row_now;
         }
 
+        if( $columns == null || $columns == 0 ) {
+		    throw new IllegalValueException( 'matrix can not have empty rows' );
+        }
+
 		$this->value = $value_now;
 	}
 
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -9,7 +9,7 @@ namespace DataValues;
 class MatrixValue extends DataValueObject {
 
 	/**
-	 * @var int|float
+	 * @var mixed
 	 */
 	private $value;
 
@@ -95,7 +95,7 @@ class MatrixValue extends DataValueObject {
 	 *
 	 * @return mixed array of arrays
 	 */
-	public function getArray() {
+	public function getMatrix() {
 		return $this->value;
 	}
 
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -55,6 +55,7 @@ class MatrixValue extends DataValueObject {
                     if( !preg_match( '/^[+-]/', $element ) ) {
                         $element = "+" . $element;
                     }
+                    $element = new DecimalValue( $element );
                 } else if (!($element instanceof DecimalValue)) {
                     $element = new DecimalValue( $element );
                 }
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -125,5 +125,4 @@ class MatrixValue extends DataValueObject {
 	public function getMatrix() {
 		return $this->value;
 	}
-
 }
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -23,50 +23,50 @@ class MatrixValue extends DataValueObject {
 	 * @throws IllegalValueException
 	 */
 	public function __construct( $value ) {
-        if ( !is_array( $value ) ) {
+		if ( !is_array( $value ) ) {
 			throw new IllegalValueException( '$value must be an array' );
-        }
+		}
 
-        if ( count( $value ) == 0 ) {
+		if ( count( $value ) == 0 ) {
 			throw new IllegalValueException( '$value can not be an empty array' );
-        }
+		}
 
-        $value_now = array();
+		$value_now = array();
 
-        $columns = null;
-        foreach ( $value as $row ) {
-            if( !is_array( $row ) ) {
-		    	throw new IllegalValueException( '$value must be an array of arrays' );
-            }
-            if( $columns == null ) {
-                $columns = count( $row );
-            } else if( $columns != count( $row ) ) {
-			    throw new IllegalValueException( 'all rows of $value must be of the same length' );
-            }
+		$columns = null;
+		foreach ( $value as $row ) {
+			if( !is_array( $row ) ) {
+				throw new IllegalValueException( '$value must be an array of arrays' );
+			}
+			if( $columns == null ) {
+				$columns = count( $row );
+			} else if( $columns != count( $row ) ) {
+				throw new IllegalValueException( 'all rows of $value must be of the same length' );
+			}
 
-            /**
-             * Filtering the matrix, converting int|float|string into
-             * DecimalValue.
-             */
-            $row_now = array();
-            foreach ( $row as $element ) {
-                if( is_string( $element ) ) {
-                    // Adding sign if missing
-                    if( !preg_match( '/^[+-]/', $element ) ) {
-                        $element = "+" . $element;
-                    }
-                    $element = new DecimalValue( $element );
-                } else if (!($element instanceof DecimalValue)) {
-                    $element = new DecimalValue( $element );
-                }
-                $row_now[] = $element;
-            }
-            $value_now[] = $row_now;
-        }
+			/**
+			 * Filtering the matrix, converting int|float|string into
+			 * DecimalValue.
+			 */
+			$row_now = array();
+			foreach ( $row as $element ) {
+				if( is_string( $element ) ) {
+					// Adding sign if missing
+					if( !preg_match( '/^[+-]/', $element ) ) {
+						$element = "+" . $element;
+					}
+					$element = new DecimalValue( $element );
+				} else if (!($element instanceof DecimalValue)) {
+					$element = new DecimalValue( $element );
+				}
+				$row_now[] = $element;
+			}
+			$value_now[] = $row_now;
+		}
 
-        if( $columns == null || $columns == 0 ) {
-		    throw new IllegalValueException( 'matrix can not have empty rows' );
-        }
+		if( $columns == null || $columns == 0 ) {
+			throw new IllegalValueException( 'matrix can not have empty rows' );
+		}
 
 		$this->value = $value_now;
 	}
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -90,4 +90,13 @@ class MatrixValue extends DataValueObject {
 		return $this;
 	}
 
+	/**
+	 * Returns the array of arrays.
+	 *
+	 * @return mixed array of arrays
+	 */
+	public function getArray() {
+		return $this->value;
+	}
+
 }
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -50,4 +50,36 @@ class MatrixValue extends DataValueObject {
 		$this->rows = $rows;
 		$this->columns = $columns;
 	}
+
+	/**
+	 * @see Serializable::serialize
+	 *
+	 * @return string
+	 */
+	public function serialize() {
+		return serialize( array(
+            $this->value,
+            $this->rows,
+            $this->columns,
+        ) );
+	}
+
+	/**
+	 * @see Serializable::unserialize
+	 *
+	 * @param string $value
+	 */
+	public function unserialize( $value ) {
+		list( $value, $rows, $columns ) = unserialize( $data );
+		$this->__construct( $value, $rows, $columns );
+	}
+
+	/**
+	 * @see DataValue::getType
+	 *
+	 * @return string
+	 */
+	public static function getType() {
+		return 'matrix';
+	}
 }
--- a/src/DataValues/MatrixValue.php
+++ b/src/DataValues/MatrixValue.php
@@ -7,46 +7,41 @@ namespace DataValues;
  * @author Andrius Merkys < andrius.merkys@gmail.com >
  */
 class MatrixValue extends DataValueObject {
+
 	/**
 	 * @var int|float
 	 */
 	private $value;
 
 	/**
-	 * @var int
-	 */
-	private $rows;
-
-	/**
-	 * @var int
-	 */
-	private $columns;
-
-	/**
 	 * Constructs a new MatrixValue object, representing the given value.
 	 *
-	 * @var NumberValue[] $value An array of numbers.
-	 * @var int $rows The number of rows.
-	 * @var int $columns The number of columns.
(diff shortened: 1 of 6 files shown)


src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Class representing a matrix.
+ *
+ * @licence GNU GPL v2+
+ * @author Andrius Merkys < andrius.merkys@gmail.com >
+ */
+class MatrixValue extends DataValueObject {
+
+	/**
+	 * @var mixed
```
thiemowmde: `@var array[]`

src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
+<?php
+namespace DataValues;
+
+use DataValues\DecimalValue;
+/**
+ * Class representing a matrix.
```
thiemowmde: Please be a little more specific of what that means, e.g. "implemented as an array of arrays of numbers".

src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
...
+
+	/**
+	 * @var mixed
+	 */
+	private $value;
+
+	/**
+	 * Constructs a new MatrixValue object, representing the given value.
+	 *
+	 * @var mixed $value An array of arrays of numbers.
```
thiemowmde: `@var array[]` please.

src/ValueParsers/MatrixParser.php
```diff
@@ -0,0 +1,44 @@
...
+	 * @throws ParseException
+	 */
+	protected function stringParse( $value ) {
+		if ( !is_string( $value ) ) {
+			throw new InvalidArgumentException( '$value must be a string' );
+		}
+
+		$decoded = json_decode( $value );
+		if( $decoded == null ) {
+			throw new ParseException( 'can not parse JSON' );
```
thiemowmde: This error message should not tell details about the implementation but about the value being unparsable.

src/ValueParsers/MatrixParser.php
```diff
@@ -0,0 +1,44 @@
...
+	 *
+	 * @return MatrixValue
+	 * @throws ParseException
+	 */
+	protected function stringParse( $value ) {
+		if ( !is_string( $value ) ) {
+			throw new InvalidArgumentException( '$value must be a string' );
+		}
+
+		$decoded = json_decode( $value );
```
thiemowmde: Nice trick to avoid writing an own parser. :+1:

brightbyte: JSON is not exactly a human friendly format... ValueParsers are for human input. It seems kind of pointless to even have one for DataValues that cannot conveniently be edited as a single line string.

Hm, actually, there is nothing that forces this to be single line. It could take a multi-line representation of the matrix.

thiemowmde:
A lot of effort went into this. Very impressive and very much appreciated. But I have to ask: Do you plan to use this, and where? Is there a ticket for this on our [Phabricator](http://phabricator.wikimedia.org/) issue tracker? If not, can you please create one to make our product manager aware of this?

src/ValueParsers/MatrixParser.php
thiemowmde: So? `json_decode` happily parses exactly the 1-line strings that should be accepted here, plus a little bit more, which is all checked in the value constructor. I currently fail to see what the problem with that is.

merkys: @thiemowmde thanks! :)
@brightbyte should the multi-line matrix input be implemented alongside JSON?

src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
...
+	 * @return MatrixValue
+	 */
+	public function getValue() {
+		return $this;
+	}
+
+	/**
+	 * Returns the array of arrays.
+	 *
+	 * @return mixed array of arrays
```
thiemowmde: Here is one more. ;-)

merkys:
@thiemowmde I will create a ticket on the [Phabricator](http://phabricator.wikimedia.org). I plan to use ``MatrixValue`` on own [Wikibase](http://wikiba.se)-based DB of material properties (tensors, in particular). Also, I think, it could be useful to sync such data with [Wikidata](http://www.wikidata.org), if seamless sync would be made possible.

merkys:
Phabricator ticket [T110534](https://phabricator.wikimedia.org/T110534) opened. Not sure if I have done it right, never did it before :)

src/DataValues/MatrixValue.php
merkys: I meant 2-dimensional array. I have now clarified in the comments that currently it's 2D array of ``DecimalValue``.

src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
...
+			if( $columns == null ) {
+				$columns = count( $row );
+			} else if( $columns != count( $row ) ) {
+				throw new IllegalValueException( 'all rows of $value must be of the same length' );
+			}
+
+			/**
+			 * Filtering the matrix, converting int|float|string into
+			 * DecimalValue.
+			 */
```
Jonas-WMDE: This should become a method

src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
...
+			$row_now = array();
+			foreach ( $row as $element ) {
+				if( is_string( $element ) ) {
+					// Adding sign if missing
+					if( !preg_match( '/^[+-]/', $element ) ) {
+						$element = "+" . $element;
+					}
+					$element = new DecimalValue( $element );
+				} else if (!($element instanceof DecimalValue)) {
```
Benestar: Code style: missing spaces around brackets

src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
...
+			}
+			$value_now[] = $row_now;
+		}
+
+		if( $columns == null || $columns == 0 ) {
+			throw new IllegalValueException( 'matrix can not have empty rows' );
+		}
+
+		$this->value = $value_now;
+	}
```
Benestar: Spaces after `if` missing in this function

src/ValueParsers/MatrixParser.php
```diff
@@ -0,0 +1,44 @@
...
+		if( $decoded == null ) {
```
Benestar: space after `if`

src/ValueParsers/MatrixParser.php
Benestar: Personally I also like the idea of using a simple json array structure as input. If you take a look at thiemo's example, it's perfectly human readable.

tests/DataValues/MatrixValueTest.php
```diff
@@ -0,0 +1,48 @@
+<?php
+namespace DataValues\Tests;
+
+use DataValues\DecimalValue;
+use DataValues\MatrixValue;
+/**
```
Benestar: new line before `/**`

tests/DataValues/MatrixValueTest.php
```diff
@@ -0,0 +1,48 @@
...
+ * @covers DataValues\MatrixValue
+ *
+ * @group DataValue
+ * @group DataValueExtensions
+ *
+ * @licence GNU GPL v2+
+ * @author Andrius Merkys < andrius.merkys@gmail.com >
+ */
+class MatrixValueTest extends DataValueTest {
+	/**
```
Benestar: as above

Benestar:
Did some nitpicking on the code style, sorry for that ;-)

Very cool implementation of a Matrix data type, I appreciate it :+1:

src/DataValues/MatrixValue.php
JeroenDeDauw: The comment as is falls into the "noise comment" category. You can remove it

src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
...
+ */
+class MatrixValue extends DataValueObject {
+
+	/**
+	 * @var array[]
+	 */
+	private $value;
+
+	/**
+	 * Constructs a new MatrixValue object, representing the given value.
```
JeroenDeDauw: The first line here is noise

src/DataValues/MatrixValue.php
```diff
@@ -0,0 +1,128 @@
...
+	public function __construct( $value ) {
+		if ( !is_array( $value ) ) {
+			throw new IllegalValueException( '$value must be an array' );
+		}
+
+		if ( count( $value ) == 0 ) {
+			throw new IllegalValueException( '$value can not be an empty array' );
+		}
+
+		$value_now = array();
```
JeroenDeDauw: We use `camelCaps`

src/DataValues/MatrixValue.php
JeroenDeDauw: Indeed. Perhaps we should add our PHPCS and PHPMD rules to this repo :)

src/DataValues/MatrixValue.php
JeroenDeDauw: Yeah definitely :)

src/ValueFormatters/MatrixFormatter.php
```diff
@@ -0,0 +1,43 @@
...
+		$raw_rows = array();
+		foreach( $value->getMatrix() as $row ) {
+			$raw_row = array();
+			foreach( $row as $element ) {
+				$raw_row[] = preg_replace( '/^\+/', '', $element->getValue() );
+			}
+			$raw_rows[] = '[' . implode( ',', $raw_row ) . ']';
+		}
+
+		$formatted = '[' . implode( ',', $raw_rows ) . ']';
```
JeroenDeDauw: If you use json_deocde at the other end, when why not json_encode here?

JeroenDeDauw:
Thanks for the contribution @merkys!

src/DataValues/MatrixValue.php
merkys: Could you please explain it? The check for the matrix input be placed in a separate method, am I right?

src/DataValues/MatrixValue.php
merkys: Sorry, I can not find the exact line to which your comment applies.

src/ValueFormatters/MatrixFormatter.php
merkys: Good idea, I will implement it!

src/DataValues/MatrixValue.php
Benestar: https://en.wikipedia.org/wiki/CamelCase

src/DataValues/MatrixValue.php
Benestar: I even think this method should be split into several small methods. Just try to keep the complexity within one method very low.

JeroenDeDauw: As a general rule, try to avoid nesting control structures or having multiple sequentially when they each need/introduce different local state. It often makes sense to do one of these regardless, however if you violate that guideline several times in a single method, it is very likely you went too far. Try to have all code directly in a method be on the same level of abstraction, one below that of the method itself.

addshore:
So this is documented as "2-dimensional array of numbers (int|float|string|DecimalValue)"

I have a friend (@tarrow) that has a usecase for a multidimensional array, containing other values.
eg.
EntityId, 1, 2, 3
EntityId, 1.2, 6.6, 9
EntityId, 0, 0, 3
etc

With this in mind would it be an idea to create some sort of abstract matrix type (that basically has no limits) and then implement a version of that for each use case?

src/ValueParsers/MatrixParser.php
brightbyte: JSON seems cumbersome for human input/output. I recommend CSV:

  1, 2
  3, 4

There is plenty of code for CSV out there, so you don't have to roll your own.

brightbyte:
I agree with @addshore: if we implement a matrix type, it should work for any size and shape of matrix.

Jonas-WMDE:
+1  any size and shape of matrix

lydiapintscher:
Folks ;-)
Let's first discuss if we want it (and in this way) before we make @merkys fix style issues please. Let's try to discuss this tomorrow.

src/ValueParsers/MatrixParser.php
merkys: It should be possible to use both JSON and CSV as input format: if the input contains square bracket, we parse it as JSON, if not -- as CSV.

src/DataValues/MatrixValue.php
merkys: OK, I've got the point.

addshore:
> Let's first discuss if we want it (and in this way) before we make @merkys fix style issues please.

+1 to that

 > I agree with @addshore: if we implement a matrix type, it should work for any size and shape of matrix.

Most of this code would probably only need a few tweaks to make this possible, but it should probably belong somewhere else (ie. not in Number)

But on the whole I really like this, and I am sure @tarrow will throw some comments in at some point!

We will get back to you after we have discussed it! :)

tarrow:
I also agree; I think it makes sense to have any size or shape and well as
flexibility as to what datatype is is in each 'cell'. i.e. not just Number.

thiemowmde:
I think the new data type in this patch can be reviewed and added as it is. We can introduce a generic Matrix datatype later and make the numeric one extend the generic one. If we take this into account while reviewing this pull request here I do not see a big problem that should block working on this. Let's keep it simple.

JeroenDeDauw:
@merkys @tarrow I'm curious what your respective usercases are. Most importantly, are they Wikibase specific, and if so, are they Wikidata specific as well?

> With this in mind would it be an idea to create some sort of abstract matrix type (that basically has no limits) and then implement a version of that for each use case?

Not clear to me if what you are talking about here is aimed at introducing an abstract type or is an implementation approach that just happens to create such a type. In case of the former, what would be the use of this type? In case of the later I'm going to scream "code reuse by inheritance - aaaaaaah!" :)

tarrow:
So I want to store the positions of objects in space where these objects
are also wikidata items. For example the structure of molecules.

I think the case for a simple numerical matrix are the matrices for
symmetry groups or stress tensors etc.. I think @merkys mentioned this
somewhere but perhaps not in this thread.

merkys:
> @merkys @tarrow I'm curious what your respective usercases are. Most importantly, are they Wikibase specific, and if so, are they Wikidata specific as well?

I see 2D matrix as another primitive data type, just as date, quantity and geo position. So MatrixData should not be limited to Wikidata only (as I am not aware of usage of DataValues outside Wikibase, I am not able to comment on the usage of MatrixData outside Wikibase).

For other uses, such as atoms in molecules, as mentioned by @tarrow, I would propose different, specific data type, since, IMO, to describe the positions of atoms in space, matrix (=unordered set of entries ``[ atom_type, x, y, z ]``) is more an implementation than the type itself.

> Not clear to me if what you are talking about here is aimed at introducing an abstract type or is an implementation approach that just happens to create such a type. In case of the former, what would be the use of this type? In case of the later I'm going to scream "code reuse by inheritance - aaaaaaah!" :)

This is supposed to be an implementation of 2D numeric matrix, and, in turn, can be made to extend an abstract matrix (=multidimensional array) type, just as stated by @thiemowmde. There's a lot of operators defined for numerical matrices that make no sense when used with other sizes or other shapes of matrices.

A pair of usecases:
* Symmetry operators (inversion, for example):
```
[
    [ -1,  0,  0 ],
    [  0, -1,  0 ],
    [  0,  0, -1 ]
]
```
* Stiffness tensor (silicon as in www.kaajakari.net/~ville/research/tutorials/elasticity_tutorial.pdf; in units of GPa for brevity):
```
[
    [ 166,  64,  64,   0,   0,   0 ],
    [  64, 166,  64,   0,   0,   0 ],
    [  64,  64, 166,   0,   0,   0 ],
    [   0,   0,   0,  80,   0,   0 ],
    [   0,   0,   0,   0,  80,   0 ],
    [   0,   0,   0,   0,   0,  80 ]
]
```

JeroenDeDauw:
> I see 2D matrix as another primitive data type, just as date, quantity and geo position. So MatrixData should not be limited to Wikidata only (as I am not aware of usage of DataValues outside Wikibase, I am not able to comment on the usage of MatrixData outside Wikibase).

My question is purely about your usecase, not about what the code should or should not bind to (which is already clear).

> This is supposed to be an implementation of 2D numeric matrix, and, in turn, can be made to extend an abstract matrix (=multidimensional array) type, just as stated by @thiemowmde. There's a lot of operators defined for numerical matrices that make no sense when used with other sizes or other shapes of matrices.

I'm afraid this does not answer the respective question either.

merkys:
> My question is purely about your usecase, not about what the code should or should not bind to (which is already clear).

I will try to explain. When I was searching for a engine/framework that I could use to build a crowdsourced database of material properties, I came upon Wikibase, for which the only major deficiency for my usecase is the lack of a way to store numeric matrices. Since Wikibase relies on DataValues' code, I have suggested numeric matrices here, as I have felt that at some point you might want to develop a support for them.

> I'm afraid this does not answer the respective question either.

I am sorry for my illiteracy. I am afraid this might be a case of code reuse by inheritance here :(

JeroenDeDauw:
Thanks for clarifying @merkys. Am I correct in understanding that you want this new DataValue for your own Wikibase installation as opposed to wanting it to be enabled on Wikidata.org?

merkys:
@JeroenDeDauw, that's not true. I would like to see this new data value both in my Wikibase installation and Wikidata.org. Some properties in Wikidata.org (at least in the field of physics) are not possible without a standardized matrix representation.

physikerwelt:
:+1: Nice; Would it be possible to make this compatible to the datatype mathematical expression. From a non technical perspective this could be seen as a mathematical expression which is restricted to a single element which is a matrix. "mathematical expression" uses tex as input format and MathML as export format to maximize compatibility to third party applications.
cf.  https://github.com/wikimedia/mediawiki-extensions-Math/blob/master/MathFormatter.php

PR #41, opened 2015-09-02
Add QuantityHtmlFormatter

thiemowmde:
[Bug: T110675](https://phabricator.wikimedia.org/T110675)

### DIFF
--- a/src/ValueFormatters/QuantityFormatter.php
+++ b/src/ValueFormatters/QuantityFormatter.php
@@ -8,7 +8,7 @@ use DataValues\QuantityValue;
 use InvalidArgumentException;
 
 /**
- * Formatter for quantity values
+ * Plain text formatter for quantity values.
  *
  * @since 0.1
  *
@@ -58,18 +58,18 @@ class QuantityFormatter extends ValueFormatterBase {
 	private $decimalFormatter;
 
 	/**
-	 * @var QuantityUnitFormatter
+	 * @var ValueFormatter|null
 	 */
-	private $unitFormatter;
+	private $vocabularyUriFormatter;
 
 	/**
 	 * @param DecimalFormatter|null $decimalFormatter
-	 * @param QuantityUnitFormatter|null $unitFormatter
+	 * @param ValueFormatter|null $vocabularyUriFormatter
 	 * @param FormatterOptions|null $options
 	 */
 	public function __construct(
 		DecimalFormatter $decimalFormatter = null,
-		QuantityUnitFormatter $unitFormatter = null,
+		ValueFormatter $vocabularyUriFormatter = null,
 		FormatterOptions $options = null
 	) {
 		parent::__construct( $options );
@@ -79,7 +79,7 @@ class QuantityFormatter extends ValueFormatterBase {
 		$this->defaultOption( self::OPT_APPLY_UNIT, true );
 
 		$this->decimalFormatter = $decimalFormatter ?: new DecimalFormatter( $this->options );
-		$this->unitFormatter = $unitFormatter ?: new BasicQuantityUnitFormatter();
+		$this->vocabularyUriFormatter = $vocabularyUriFormatter;
 
 		// plain composition should be sufficient
 		$this->decimalMath = new DecimalMath();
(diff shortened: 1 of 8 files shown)


brightbyte:
Bug: T111172

src/ValueFormatters/QuantityHtmlFormatter.php
```diff
@@ -0,0 +1,35 @@
...
+
+		if ( $this->vocabularyUriFormatter !== null ) {
+			$unit = $this->vocabularyUriFormatter->format( $unit );
+
+			if ( $unit === null ) {
+				return $html;
+			}
+		}
+
+		return $html . ' <span class="unit">' . htmlspecialchars( $unit ) . '</span>';
```
brightbyte: Perhaps put the unit URI into a data attribute, for completeness?

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -168,4 +181,23 @@ private function formatMargin( DecimalValue $margin, $roundingExponent ) {
...
+		if ( $this->vocabularyUriFormatter !== null ) {
+			$unit = $this->vocabularyUriFormatter->format( $unit );
+
+			if ( $unit === null ) {
+				return $number;
+			}
+		}
+
+		// TODO: localizable pattern for placement (before/after, separator)
+		return $number . ' ' . $unit;
```
brightbyte: Since this depends on the locale, it seems likely that we will want to inject the logic for combining unit and number.

brightbyte:
+1 Some questions, but no major issues.

src/ValueFormatters/QuantityHtmlFormatter.php
thiemowmde: * Data attribute: Could do, but why? Not relevant for the things we want to do in the sprint. YAGNI.
* A "configurable" CSS class name? Who, no. The fact that it's "generic" is intended. Code that applies styles on this without a proper selector is wrong anyway. But I agree that we should change this to "wb-unit".

src/ValueFormatters/QuantityFormatter.php
thiemowmde: True. I think this is what we tried with the "unit appender interface" you called `QuantityUnitFormatter`. I would like to get rid of this and start over when we really need it.

src/ValueFormatters/QuantityHtmlFormatter.php
brightbyte: The problem with the generic class name is that it would be present elsewhere in the DOM, which would make it harder to find the "correct" element for the unit.

src/ValueFormatters/QuantityHtmlFormatter.php
```diff
@@ -0,0 +1,35 @@
...
+class QuantityHtmlFormatter extends QuantityFormatter {
+
+	/**
+	 * @param string $number
+	 * @param string $unit
+	 *
+	 * @return string HTML
+	 */
+	protected function applyUnit( $number, $unit ) {
+		$html = htmlspecialchars( $number );
```
brightbyte: So, if applyUnit() is not called (because self::OPT_APPLY_UNIT is false, or no unit is given), this formatter would return $number unescaped? Not good, who knows what it may contain after localization...

src/ValueFormatters/QuantityHtmlFormatter.php
thiemowmde: That's what CSS selectors are for. You can combine them.

I'm not sure to what "knowledge" you refer. This is a string. I can call it "nsa-unit", this doesn't imply knowledge about the NSA.

src/ValueFormatters/QuantityHtmlFormatter.php
thiemowmde: No. Please read the code again.

thiemowmde: Dang, now I can see it. You are right. -2.

brightbyte: How about requiering the NumberFormatter etc to already return HTML?
And how about having a single method for composing all the bits, where each bit should already be HTML?

PR #43, opened 2015-09-04
Add more rounding tests for QuantityFormatter

thiemowmde:
~~I'm aware that this is incomplete and you can come up with an endless amount of examples where the rounding still acts unexpected and/or still does round to much or to less. Actually, please post your examples (input and expected output) and I will add them.~~

[Bug: T95425](https://phabricator.wikimedia.org/T95425)


thiemowmde:
I rebased and reworked this patch. It became *almost* obsolete with #68. Now it does nothing any more except adding a few more test cases. I still find these useful.

PR #44, opened 2015-09-07
Add $quantityWithUnitFormat parameter to Quantity[Html]Formatter

thiemowmde:
The QuantityFormatter constructor does have a breaking change anyway. I'm reordering the parameters so that options are first, then the two formatters for the number and the unit, then the format string to combine the two.

[Bug: T111186](http://phabricator.wikimedia.org/T111186)

### DIFF
--- a/src/ValueFormatters/QuantityFormatter.php
+++ b/src/ValueFormatters/QuantityFormatter.php
@@ -63,16 +63,26 @@ class QuantityFormatter extends ValueFormatterBase {
 	private $vocabularyUriFormatter;
 
 	/**
+	 * @var string
+	 */
+	private $quantityWithUnitFormat;
+
+	/**
 	 * @since 0.6
 	 *
 	 * @param FormatterOptions|null $options
 	 * @param DecimalFormatter|null $decimalFormatter
 	 * @param ValueFormatter|null $vocabularyUriFormatter
+	 * @param string|null $quantityWithUnitFormat Format string with two placeholders, $1 for the
+	 * number and $2 for the unit. Warning, this must be under the control of the application, not
+	 * under the control of the user, because it allows HTML injections in subclasses that return
+	 * HTML.
 	 */
 	public function __construct(
 		FormatterOptions $options = null,
 		DecimalFormatter $decimalFormatter = null,
-		ValueFormatter $vocabularyUriFormatter = null
+		ValueFormatter $vocabularyUriFormatter = null,
+		$quantityWithUnitFormat = null
 	) {
 		parent::__construct( $options );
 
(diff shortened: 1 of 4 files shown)


src/ValueFormatters/QuantityFormatter.php
```diff
@@ -63,14 +63,25 @@ class QuantityFormatter extends ValueFormatterBase {
...
+	/**
+	 * @since 0.6
+	 *
+	 * @param FormatterOptions|null $options
 	 * @param DecimalFormatter|null $decimalFormatter
 	 * @param ValueFormatter|null $vocabularyUriFormatter
-	 * @param FormatterOptions|null $options
+	 * @param string|null $quantityWithUnitFormat Warning, this must be under the control of the
+	 * application, not under the control of the user, because it allows HTML injections in
+	 * subclasses that return HTML.
```
brightbyte: Please describe the expected format of this string

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -80,12 +91,22 @@ public function __construct(
...
 		// plain composition should be sufficient
 		$this->decimalMath = new DecimalMath();
 	}
 
 	/**
+	 * @since 0.6
+	 *
+	 * @return string
+	 */
+	protected function getQuantityWithUnitFormat() {
```
brightbyte: Why do we need this method? Why is it protected? What can be achieved by overriding it?

brightbyte:
Seems sensible, though not really needed right now.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -63,14 +63,25 @@ class QuantityFormatter extends ValueFormatterBase {
...
+	 * @param FormatterOptions|null $options
 	 * @param DecimalFormatter|null $decimalFormatter
 	 * @param ValueFormatter|null $vocabularyUriFormatter
-	 * @param FormatterOptions|null $options
+	 * @param string|null $quantityWithUnitFormat Warning, this must be under the control of the
+	 * application, not under the control of the user, because it allows HTML injections in
+	 * subclasses that return HTML.
 	 */
 	public function __construct(
+		FormatterOptions $options = null,
```
brightbyte: Changing the order here makes this a breaking change. 
Without this, we could push this until after the major release.

thiemowmde: No. It already is a breaking change.

brightbyte: How? Did I miss something?

thiemowmde: You merged it: https://github.com/DataValues/Number/pull/41/files#diff-8c77e0fb91ce0037fa8e2fd7b88a35ab

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -63,14 +63,26 @@ class QuantityFormatter extends ValueFormatterBase {
...
 	 * @param DecimalFormatter|null $decimalFormatter
 	 * @param ValueFormatter|null $vocabularyUriFormatter
-	 * @param FormatterOptions|null $options
+	 * @param string|null $quantityWithUnitFormat Format string with two placeholders, $1 for the
+	 * number and $2 for the unit. Warning, this must be under the control of the application, not
+	 * under the control of the user, because it allows HTML injections in subclasses that return
+	 * HTML.
 	 */
 	public function __construct(
+		FormatterOptions $options = null,
```
thiemowmde: I have split this change off to #45. Please merge #45 first and let me know, I will rebase #44 then.

brightbyte:
+2 from me. I would merge this (no need for #45 really), but travis sais no :(

thiemowmde:
Where does Travis say no?

thiemowmde:
Can you please merge #45?

thiemowmde:
Rebased and reworked QuantityHtmlFormatter a bit to not duplicate the "quantityWithUnitFormat" code in QuantityFormatter::formatQuantityValue, instead it overrides formatNumber and formatUnit now.

PR #46, opened 2015-09-09
Drop underspecified and incomplete unit parsing

thiemowmde:
This is an alternative approach to unblock #37. The unit parsing is broken. Either fix it or remove it.

**PS:** This does not mean #37 becomes invalid. Both should be merged.

### DIFF
--- a/tests/ValueParsers/QuantityParserTest.php
+++ b/tests/ValueParsers/QuantityParserTest.php
@@ -59,11 +59,11 @@ class QuantityParserTest extends StringValueParserTest {
 
 			'1.4e-2' => QuantityValue::newFromNumber( '+0.014', '1', '+0.015', '+0.013' ),
 			'1.4e3' => QuantityValue::newFromNumber( '+1400', '1', '+1500', '+1300' ),
-			'1.4e3!m' => QuantityValue::newFromNumber( '+1400', 'm', '+1400', '+1400' ),
-			'1.4e3m2' => QuantityValue::newFromNumber( '+1400', 'm2', '+1500', '+1300' ),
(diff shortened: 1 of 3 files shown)


tests/ValueParsers/QuantityParserTest.php
```diff
@@ -95,12 +95,12 @@ public function validInputProvider() {
 			'5.3 ±-0.2' => QuantityValue::newFromNumber( '+5.3', '1', '+5.5', '+5.1' ),
 
 			// units
-			'5.3+-0.2cm' => QuantityValue::newFromNumber( '+5.3', 'cm', '+5.5', '+5.1' ),
-			'10.003! km' => QuantityValue::newFromNumber( '+10.003', 'km', '+10.003', '+10.003' ),
-			'-200~ %  ' => QuantityValue::newFromNumber( -200, '%', -199, -201 ),
-			'100003 m³' => QuantityValue::newFromNumber( 100003, 'm³', 100004, 100002 ),
-			'3.±-0.2µ' => QuantityValue::newFromNumber( '+3', 'µ', '+3.2', '+2.8' ),
-			'+00.20 Å' => QuantityValue::newFromNumber( '+0.20', 'Å', '+0.21', '+0.19' ),
+			'5.4+-0.2' => QuantityValue::newFromNumber( '+5.4', '1', '+5.6', '+5.2' ),
```
brightbyte: Rather drop this section, it doesn't make sense if no units are accepted.

brightbyte:
I agree with removing unit parsing.

Rationale: The current implementation for parsing units in quantities hard-codes an ad-hoc regex and has no mechanism for normalizing (unlocalizing) the unit. It also does not allow applications to specify the syntax to use for units.

Since Wikibase doesn't currently need parsing units from a quantity string, it's ok to drop this functionality until we need it, and then implement it according to our needs.

PR #52, opened 2015-12-28
Remove deprecated QuantityUnitFormatter

thiemowmde:
This was introduced in #31 for https://gerrit.wikimedia.org/r/216713 but became unused some weeks later with #41 and https://gerrit.wikimedia.org/r/235495. This here is not actually a "formatter" but an "appender" with a very, very limited use.


mariushoch:
Deprecated in 0.6.0, confirmed unused.

PR #55, opened 2016-02-23
[WIP] Allow QuantityValue::upper/lowerBounds to be null

thiemowmde:
* ~~Tests missing.~~
* ~~Some method will still explode when the bounds are null.~~
* ~~Formatter and parser must be adopted.~~

Breaking because users of this class don't expect this to be null.

[Bug: T115270](https://phabricator.wikimedia.org/T115270)
[Bug: T133042](https://phabricator.wikimedia.org/T133042)

### DIFF
--- a/src/DataValues/QuantityValue.php
+++ b/src/DataValues/QuantityValue.php
@@ -14,6 +14,7 @@ use InvalidArgumentException;
  *
  * @license GPL-2.0+
  * @author Daniel Kinzler
+ * @author Thiemo Mättig
  */
 class QuantityValue extends DataValueObject {
 
@@ -34,14 +35,14 @@ class QuantityValue extends DataValueObject {
 	/**
 	 * The quantity's upper bound
 	 *
-	 * @var DecimalValue
+	 * @var DecimalValue|null
 	 */
 	private $upperBound;
 
 	/**
 	 * The quantity's lower bound
 	 *
-	 * @var DecimalValue
+	 * @var DecimalValue|null
 	 */
 	private $lowerBound;
 
@@ -52,17 +53,26 @@ class QuantityValue extends DataValueObject {
 	 *
 	 * @param DecimalValue $amount
 	 * @param string $unit A unit identifier. Must not be empty, use "1" for unit-less quantities.
-	 * @param DecimalValue $upperBound The upper bound of the quantity, inclusive.
-	 * @param DecimalValue $lowerBound The lower bound of the quantity, inclusive.
+	 * @param DecimalValue|null $upperBound The upper bound of the quantity, inclusive.
+	 * @param DecimalValue|null $lowerBound The lower bound of the quantity, inclusive.
 	 *
 	 * @throws IllegalValueException
 	 */
-	public function __construct( DecimalValue $amount, $unit, DecimalValue $upperBound, DecimalValue $lowerBound ) {
-		if ( $lowerBound->compare( $amount ) > 0 ) {
+	public function __construct(
+		DecimalValue $amount,
+		$unit,
+		DecimalValue $upperBound = null,
+		DecimalValue $lowerBound = null
+	) {
+		if ( is_null( $upperBound ) !== is_null( $lowerBound ) ) {
+			throw new IllegalValueException( '$upperBound and $lowerBound must both be defined or both undefined' );
+		}
+
+		if ( $lowerBound && $lowerBound->compare( $amount ) > 0 ) {
 			throw new IllegalValueException( '$lowerBound ' . $lowerBound->getValue() . ' must be <= $amount ' . $amount->getValue() );
 		}
 
-		if ( $upperBound->compare( $amount ) < 0 ) {
+		if ( $upperBound && $upperBound->compare( $amount ) < 0 ) {
 			throw new IllegalValueException( '$upperBound ' . $upperBound->getValue() . ' must be >= $amount ' . $amount->getValue() );
 		}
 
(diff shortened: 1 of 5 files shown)


mariushoch:
Ticket? What is `null` supposed to mean?

thiemowmde:
We just mentioned this in the story time. We must support quantities where the original source did not stated any precision. Not sure what the ticket is. As said, WIP.

src/DataValues/QuantityValue.php
```diff
@@ -283,13 +292,17 @@ public function getUncertainty() {
 	 *
 	 * @since 0.1
 	 *
-	 * @return DecimalValue
+	 * @return DecimalValue|null
 	 */
 	public function getUncertaintyMargin() {
+		if ( $this->upperBound === null ) {
```
brightbyte: pleas add a hasBounds() method

brightbyte:
Looks good to me, but I didn't test. Thiemo, what do you think is missing here?

thiemowmde:
What's missing:
* GitHub responding again.
* A manual rebase.
* Tests, tests, tests. All I did for now is adjusting existing test cases, but I did not added any new tests for the new "null precision" feature.

src/DataValues/QuantityValue.php
brightbyte: Or perhaps we could move bounds to a subclass, BoundedQuantity, instead of checking for bounds in every method.

src/DataValues/QuantityValue.php
```diff
@@ -52,17 +53,26 @@ class QuantityValue extends DataValueObject {
...
-	 * @param DecimalValue $lowerBound The lower bound of the quantity, inclusive.
+	 * @param DecimalValue|null $upperBound The upper bound of the quantity, inclusive.
+	 * @param DecimalValue|null $lowerBound The lower bound of the quantity, inclusive.
 	 *
 	 * @throws IllegalValueException
 	 */
-	public function __construct( DecimalValue $amount, $unit, DecimalValue $upperBound, DecimalValue $lowerBound ) {
-		if ( $lowerBound->compare( $amount ) > 0 ) {
+	public function __construct(
+		DecimalValue $amount,
```
brightbyte: unrelated: we should rename that to "nominalValue"

src/DataValues/QuantityValue.php
```diff
@@ -52,17 +53,26 @@ class QuantityValue extends DataValueObject {
...
+		$unit,
+		DecimalValue $upperBound = null,
+		DecimalValue $lowerBound = null
+	) {
+		if ( is_null( $upperBound ) !== is_null( $lowerBound ) ) {
+			throw new IllegalValueException( '$upperBound and $lowerBound must both be defined or undefined' );
```
brightbyte: ..."must either both be defined or both be undefined".

src/DataValues/QuantityValue.php
```diff
@@ -344,14 +361,18 @@ public function getOrderOfUncertainty() {
...
+	 * @return int|null
 	 */
 	public function getSignificantFigures() {
 		$math = new DecimalMath();
 
 		// $orderOfUncertainty is +/- 200 -> 2; +/- 0.02 -> -2
 		$orderOfUncertainty = $this->getOrderOfUncertainty();
 
+		if ( $orderOfUncertainty === null ) {
+			return null;
```
brightbyte: If no uncertainty is given, all figures should be considered significant. There is no need to return null here.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -191,13 +191,13 @@ private function getRoundingExponent( QuantityValue $quantity ) {
...
-	 * @param DecimalValue $margin
+	 * @param DecimalValue|null $margin
 	 * @param int $roundingExponent
 	 *
 	 * @return string|null Text
 	 */
-	private function formatMargin( DecimalValue $margin, $roundingExponent ) {
-		if ( $this->options->getOption( self::OPT_SHOW_UNCERTAINTY_MARGIN ) ) {
+	private function formatMargin( DecimalValue $margin = null, $roundingExponent ) {
+		if ( $margin !== null && $this->options->getOption( self::OPT_SHOW_UNCERTAINTY_MARGIN ) ) {
```
brightbyte: for backwards compat, we need multiple values now: NEVER, IF_KNOWN, and IF_NOT_ZERO. IF_KNOWN is what we really want, IF_NOT_ZERO can be used to emulate the current behavior.

tests/ValueFormatters/QuantityHtmlFormatterTest.php
```diff
@@ -65,6 +65,10 @@ private function getQuantityHtmlFormatter(
 	 */
 	public function validProvider() {
 		return array(
+			'Unknown uncertainty' => array(
```
brightbyte: i like unknown uncertainty ;)

src/ValueFormatters/QuantityFormatter.php
thiemowmde: I don't think this gives us anything. How would that even be rendered? `1±?`?

brightbyte: IF_NOT_ZERO means IF_KNOWN_AND_NOT_ZERO, of course. basically, we'll want an option to show bounds when given (including zero), or to always hide the bounds if they are zero (current behavior).

JonasKress:
@brightbyte What is needed to get this merged?
+1 from my side

thiemowmde:
This proposal became obsolete with #76.

PR #58, opened 2016-04-21
T110728: Trim trailing newlines from DecimalValues

thiemowmde:
This will make bogus QuantityValues in the database magically disappear over time, every time an entity is edited. The newlines will also immediately disappear from all exports.

[Bug: T110728](https://phabricator.wikimedia.org/T110728)

### DIFF
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -47,38 +47,30 @@ class DecimalValue extends DataValueObject {
 	 *
 	 * @param string|int|float $value If given as a string, the value must match
 	 *                         QUANTITY_VALUE_PATTERN.
+	 *
+	 * @throws IllegalValueException
 	 */
 	public function __construct( $value ) {
 		if ( is_int( $value ) || is_float( $value ) ) {
 			$value = $this->convertToDecimal( $value );
-		}
-
-		$this->assertNumberString( $value );
-
-		// make "negative" zero positive
-		$value = preg_replace( '/^-(0+(\.0+)?)$/', '+\1', $value );
-
-		$this->value = $value;
-	}
-
-	/**
-	 * Checks that the given value is a number string.
-	 *
-	 * @param string $number The value to check
-	 *
-	 * @throws IllegalValueException
-	 */
-	private function assertNumberString( $number ) {
-		if ( !is_string( $number ) ) {
+		} elseif ( !is_string( $value ) ) {
 			throw new IllegalValueException( '$number must be a numeric string.' );
 		}
 
-		if ( strlen( $number ) > 127 ) {
+		$value = trim( $value );
+
+		if ( strlen( $value ) > 127 ) {
 			throw new IllegalValueException( 'Value must be at most 127 characters long.' );
 		}
-
-		if ( !preg_match( self::QUANTITY_VALUE_PATTERN, $number ) ) {
+		if ( !preg_match( self::QUANTITY_VALUE_PATTERN, $value ) ) {
 			throw new IllegalValueException( 'Value must match the pattern for decimal values.' );
+		}
+
+		$this->value = $value;
+
+		// make "negative" zero positive
+		if ( $this->isZero() ) {
+			$this->value = '+' . substr( $this->value, 1 );
 		}
 	}
 
@@ -93,10 +85,6 @@ class DecimalValue extends DataValueObject {
 	 * @throws InvalidArgumentException
 	 */
 	private function convertToDecimal( $number ) {
-		if ( !is_int( $number ) && !is_float( $number ) ) {
-			throw new InvalidArgumentException( '$number must be an int or float.' );
-		}
-
 		if ( $number === NAN || abs( $number ) === INF ) {
 			throw new InvalidArgumentException( '$number must not be NAN or INF.' );
 		}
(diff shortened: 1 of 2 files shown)


src/DataValues/DecimalValue.php
```diff
@@ -138,8 +122,8 @@ public function compare( self $that ) {
 			return 0;
 		}
 
-		$a = $this->getValue();
-		$b = $that->getValue();
+		$a = $this->value;
+		$b = $that->value;
```
brightbyte: this kind of "btw" change makes review harder for me.

src/DataValues/DecimalValue.php
```diff
@@ -93,33 +85,25 @@ private function assertNumberString( $number ) {
 	 * @throws InvalidArgumentException
 	 */
 	private function convertToDecimal( $number ) {
-		if ( !is_int( $number ) && !is_float( $number ) ) {
-			throw new InvalidArgumentException( '$number must be an int or float.' );
-		}
```
brightbyte: Why is nit necessary to remove this constraint? I don't remember why it's here, and it my be good to remove it, but why in this patch?

src/DataValues/DecimalValue.php
```diff
@@ -47,39 +47,31 @@ class DecimalValue extends DataValueObject {
...
 
-		if ( strlen( $number ) > 127 ) {
+		$value = trim( $value );
+
+		if ( strlen( $value ) > 127 ) {
 			throw new IllegalValueException( 'Value must be at most 127 characters long.' );
 		}
-
-		if ( !preg_match( self::QUANTITY_VALUE_PATTERN, $number ) ) {
+		if ( !preg_match( self::QUANTITY_VALUE_PATTERN, $value ) ) {
```
brightbyte: QUANTITY_VALUE_PATTERN ends with $, shouldn't that be changed to \z?

src/DataValues/DecimalValue.php
```diff
@@ -93,33 +85,25 @@ private function assertNumberString( $number ) {
...
 
 			if ( $decimal[0] === '.' ) {
 				$decimal = '0' . $decimal;
 			}
 
-			$last = strlen($decimal)-1;
-
-			if ( $decimal[$last] === '.' ) {
-				$decimal = $decimal . '0';
+			if ( substr( $decimal, -1 ) === '.' ) {
```
brightbyte: I find the code more readable with a local variable. why mess with unrelated code when fixing trimming and validation?

src/DataValues/DecimalValue.php
thiemowmde: I worked extra hard to minimize the diff to an absolute minimum. I could have uploaded these 3 lines as a separate patch, and do this most of the time, but found its not worth it in this case.

brightbyte:
Seems fine overall, though I don't like the "marginal" changes much.
We should probably change QUANTITY_VALUE_PATTERN, though.

src/DataValues/DecimalValue.php
thiemowmde: We just talked about this principle for an hour. This is not unrelated. This is the same class. Boy scout rule. Cleanup code you find hard to read while you are working on closely related code. I will not stop doing this, and I find it more and more frustrating to defeat my position.

src/DataValues/DecimalValue.php
thiemowmde: I did and reverted it to minimize the diff. It does not make a difference with the trim in place.

src/DataValues/DecimalValue.php
thiemowmde: This is done two times, literally one line of code before this method is called.

src/DataValues/DecimalValue.php
brightbyte: But why change this at all?

src/DataValues/DecimalValue.php
brightbyte: Ah, right, ok.

src/DataValues/DecimalValue.php
brightbyte: True, but it's a constant that may be used elsewhere... since this patch is about properly trimming, and fixing the odd behavior of $ in pcre, it seems a good time and place to turn the $ into a \z

src/DataValues/DecimalValue.php
brightbyte: In general, I agree with cleaning up code that is "close by". But I find myself trying to avoid this more and more, as I often find that it makes review harder. I have to look a lot closer to find the "meat" of the change. 

Also, some cleanup is debatable. In this case, I find the new version of the code (slightly) harder to read.

thiemowmde: I disagree, otherwise I would not have done this. "Last" what? I can't tell easily if there is an off-by-one error in this `strlen(…)-1` and the line below where it is used.

src/DataValues/DecimalValue.php
thiemowmde: Why polluting the stack trace to access a property?

src/DataValues/DecimalValue.php
```diff
@@ -40,46 +40,38 @@ class DecimalValue extends DataValueObject {
 	 * Regular expression for matching decimal strings that conform to the format
 	 * described in the class level documentation of @see DecimalValue.
 	 */
-	const QUANTITY_VALUE_PATTERN = '/^[-+]([1-9]\d*|\d)(\.\d+)?$/';
+	const QUANTITY_VALUE_PATTERN = '/^[-+]([1-9]\d*|\d)(\.\d+)?\z/';
```
smalyshev: /^[-+]([1-9]\d*|\d)(\.\d+)?$/D should also work.

PR #64, opened 2016-05-25
Do not allow VocabularyUriFormatter to be null

thiemowmde:
Introduced in 0.6. This feature is never used, not even in tests. Was meant to be kind of an option to disable this feature, but as you can see there is an actual option for that.


thiemowmde:
@JonasKress can we squish this into the 0.8 release? It's unused, but technically a breaking change. I do not want this to sit here until 0.9 or later.

PR #66, opened 2016-06-23
Factor UnboundedQuantityValue out of QuantityValue

brightbyte:
Code experiment exploring the possibility to use subclassing to model unbounded vs bounded quantities.


src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,381 @@
...
+	 * @param string|int|float|DecimalValue|QuantityValue $number
+	 * @param string $unit A unit identifier. Must not be empty, use "1" for unit-less quantities.
+	 * @param string|int|float|DecimalValue|null $upperBound
+	 * @param string|int|float|DecimalValue|null $lowerBound
+	 *
+	 * @return self
+	 * @throws IllegalValueException
+	 */
+	public static function newFromNumber( $number, $unit = '1', $upperBound = null, $lowerBound = null ) {
+		if ( $number instanceof QuantityValue ) {
```
thiemowmde: This may already have a unit and bounds, which will be ignored. I don't think this is helpful.

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,381 @@
...
+	 *
+	 * @param string|int|float|DecimalValue|QuantityValue $number
+	 * @param string $unit
+	 * @param string|int|float|DecimalValue|null $upperBound
+	 * @param string|int|float|DecimalValue|null $lowerBound
+	 *
+	 * @return self
+	 */
+	public static function newFromDecimal( $number, $unit = '1', $upperBound = null, $lowerBound = null ) {
+		if ( $number instanceof QuantityValue ) {
```
thiemowmde: As above.

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,381 @@
...
+
+	/**
+	 * @see DataValue::getType
+	 *
+	 * @since 0.1
+	 *
+	 * @return string
+	 */
+	public static function getType() {
+		return 'quantity'; // FIXME: should this have a different type id? wehat will that break?
```
thiemowmde: I think: both is the same, the fact that we use subclasses has nothing to do with the data model. It's to make the code nicer.

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,381 @@
...
+	 * @see DataValue::getArrayValue
+	 *
+	 * @since 0.1
+	 *
+	 * @return array
+	 */
+	public function getArrayValue() {
+		return array(
+			'amount' => $this->amount->getArrayValue(),
+			'unit' => $this->unit,
```
thiemowmde: Could call the super class here.

src/DataValues/QuantityValue.php
```diff
@@ -5,9 +5,11 @@
 use InvalidArgumentException;
 
 /**
- * Class representing a quantity with associated unit and uncertainty interval.
+ * Class representing a quantity with associated unit.
  * The amount is stored as a @see DecimalValue object.
  *
+ * For quantities with a known uncertainty interval, see BoundedQuantityValue.
```
thiemowmde: Have an actual `@see`?

src/DataValues/QuantityValue.php
```diff
@@ -93,36 +67,32 @@ public function __construct( DecimalValue $amount, $unit, DecimalValue $upperBou
...
+		$number = self::asDecimalValue( 'amount', $number );
 
-		return new self( $amount, $unit, $upperBound, $lowerBound );
+		return new self( $number, $unit );
 	}
 
 	/**
 	 * @see newFromNumber
 	 *
 	 * @deprecated since 0.1, use newFromNumber instead
```
thiemowmde: Breaking change anyway, can easily be removed now.

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,381 @@
...
+		if ( $precision === 0.0 ) {
+			// If there is no uncertainty, the order of uncertainty is a bit more than what we have digits for.
+			return -strlen( $this->amount->getFractionalPart() );
+		}
+
+		// e.g. +/- 200 -> 2; +/- 0.02 -> -2
+		// note: we really want floor( log10( $precision ) ), but have to account for
+		// small errors made in the floating point operations above.
+		// @todo: use bcmath (via DecimalMath) to avoid this if possible
+		$orderOfUncertainty = floor( log10( $precision + 0.0000000005 ) );
```
thiemowmde: Let's please not re-introduce this odd code. Let's rewrite this where it is needed.

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,381 @@
...
+	 * Returns the number of significant figures in the amount-string,
+	 * counting the decimal point, but not counting the leading sign.
+	 *
+	 * Note that this calculation assumes a symmetric uncertainty interval, and can be misleading
+	 *
+	 * @since 0.1
+	 *
+	 * @return int
+	 */
+	public function getSignificantFigures() {
```
thiemowmde: Not used anywhere, can simply be removed.

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,381 @@
...
+	 *        The first parameter passed to $transformation is the DecimalValue to transform
+	 *        In addition, any extra parameters passed to transform() will be passed through
+	 *        to the transformation callback.
+	 *
+	 * @param mixed ... Any extra parameters will be passed to the $transformation function.
+	 *
+	 * @throws InvalidArgumentException
+	 * @return self
+	 */
+	public function transform( $newUnit, $transformation ) {
```
thiemowmde: Not used anywhere. Does not belong to a value object. Please remove.

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,381 @@
...
+	 *
+	 * @param mixed $data
+	 *
+	 * @return self
+	 * @throws IllegalValueException
+	 */
+	public static function newFromArray( $data ) {
+		self::requireArrayFields( $data, array( 'amount', 'unit', 'upperBound', 'lowerBound' ) );
+
+		return new static(
```
thiemowmde: This could call the parent class when the two fields are missing (for convenience).

src/DataValues/QuantityValue.php
```diff
@@ -237,128 +202,6 @@ public function getAmount() {
...
-	 * Returns the number of significant figures in the amount-string,
-	 * counting the decimal point, but not counting the leading sign.
-	 *
-	 * Note that this calculation assumes a symmetric uncertainty interval, and can be misleading
-	 *
-	 * @since 0.1
-	 *
-	 * @return int
-	 */
-	public function getSignificantFigures() {
```
brightbyte: I'm happy to discuss factoring the significant digits stuff out, but let's please do that in a later pull request, so we keep steps small and don't bock one change because of discussion about the other.
I agree that it should target the same major release, though.

src/DataValues/BoundedQuantityValue.php
brightbyte: Let's please not rewrite this odd code, it was rather tricky to get right; we can factor it out into a separate class, but let's do that in a separate change.

Let's also keep in mind that the transform() method is central to unit conversion. Even if it's not really used at the moment, we should keep it around.

src/DataValues/BoundedQuantityValue.php
brightbyte: We will need it for unit conversion, for formatting and indexing. Deleting tricky code is asking for trouble. I'm open to moving it though, see above.

src/DataValues/BoundedQuantityValue.php
brightbyte: I prefer not to, to keep this symmetrical and consistent with newFromArray

src/DataValues/QuantityValue.php
brightbyte: what can be removed?

composer.json
```diff
@@ -25,6 +25,10 @@
 		"data-values/interfaces": "~0.2.0|~0.1.5",
 		"data-values/common": "~0.3.0|~0.2.0"
 	},
+	"require-dev": {
+		"mediawiki/mediawiki-codesniffer": "0.7.2",
+		"phpunit/phpunit": "4.8.24"
```
JeroenDeDauw: Why disallow newer versions?

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,352 @@
...
+ * The amount is stored as a @see DecimalValue object.
+ *
+ * @see QuantityValue For quantities with no known uncertainty interval.
+ * For simple numeric amounts use @see NumberValue.
+ *
+ * @note BoundedQuantityValue and plain QuantityValue both use the value type ID "quantity".
+ * The fact that we use subclassing to model the bounded vs the unbounded case should be
+ * considered an implementation detail.
+ *
+ * @since 0.1
```
JeroenDeDauw: ?

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,352 @@
...
+	 * The quantity's lower bound
+	 *
+	 * @var DecimalValue
+	 */
+	private $lowerBound;
+
+	/**
+	 * Constructs a new QuantityValue object, representing the given value.
+	 *
+	 * @since 0.1
```
JeroenDeDauw: ?

src/DataValues/BoundedQuantityValue.php
```diff
@@ -0,0 +1,352 @@
...
+
+	/**
+	 * The quantity's lower bound
+	 *
+	 * @var DecimalValue
+	 */
+	private $lowerBound;
+
+	/**
+	 * Constructs a new QuantityValue object, representing the given value.
```
JeroenDeDauw: 42be4aef-36a8-a182-1851172

composer.json
brightbyte: This is just copy&paste from core. I don't care much.

src/DataValues/BoundedQuantityValue.php
brightbyte: Omnomnom, copy pasta!

composer.json
JeroenDeDauw: Copy pasting from MediaWiki does not strike me as a good idea :)

composer.json
```diff
@@ -25,6 +25,9 @@
 		"data-values/interfaces": "~0.2.0|~0.1.5",
 		"data-values/common": "~0.3.0|~0.2.0"
 	},
+	"require-dev": {
+		"phpunit/phpunit": "~4.8"
```
JeroenDeDauw: Why not the current major version of PHPUnit?

brightbyte: I prefer for all our libraries to use the same version of tools and libraries. Otherwise, it becomes confusing what feature you can use where. If we want to move to 5.x, let's do that for all our stuff.

src/DataValues/QuantityValue.php
brightbyte: oh, the deprecated method, right.

src/DataValues/QuantityValue.php
```diff
@@ -8,28 +8,19 @@
  * Class representing a quantity with associated unit and uncertainty interval.
  * The amount is stored as a @see DecimalValue object.
  *
+ * @see UnboundedQuantityValue For quantities with no known uncertainty interval.
  * For simple numeric amounts use @see NumberValue.
  *
- * @since 0.1
```
thiemowmde: Did the fact that the class exists since 0.1 changed?

composer.json
thiemowmde: Should be a separate patch, please.

src/DataValues/QuantityValue.php
```diff
@@ -91,155 +70,59 @@ public function __construct( DecimalValue $amount, $unit, DecimalValue $upperBou
 	 * @note: if the amount or a bound is given as a string, the string must conform
 	 * to the rules defined by @see DecimalValue.
 	 *
-	 * @since 0.1
+	 * @since 0.7
```
thiemowmde: As with all `@since` tags in this file, this should not change because the method signature did not changed, as far as I can see. That was the whole point of renaming the two classes, wasn't it?

src/DataValues/QuantityValue.php
```diff
@@ -91,155 +70,59 @@ public function __construct( DecimalValue $amount, $unit, DecimalValue $upperBou
...
 	 *
-	 * @param string|int|float|DecimalValue $amount
+	 * @param string|int|float|DecimalValue|UnboundedQuantityValue $number
```
thiemowmde: I believe this is a mistake. We talked about this and decided to not allow using an other QuantityValue as a number because you can pass unit and bounds twice that way.

Same a few lines below.

src/DataValues/QuantityValue.php
```diff
@@ -91,155 +70,59 @@ public function __construct( DecimalValue $amount, $unit, DecimalValue $upperBou
...
 	 * @param string $unit A unit identifier. Must not be empty, use "1" for unit-less quantities.
 	 * @param string|int|float|DecimalValue|null $upperBound
 	 * @param string|int|float|DecimalValue|null $lowerBound
 	 *
-	 * @return self
+	 * @return QuantityValue
```
thiemowmde: A mistake that happened when renaming the files? Should return `self`.

src/DataValues/QuantityValue.php
```diff
@@ -91,155 +70,59 @@ public function __construct( DecimalValue $amount, $unit, DecimalValue $upperBou
...
-	 * @since 0.1
+	 * @since 0.7
 	 *
 	 * @param string $data
 	 */
 	public function unserialize( $data ) {
 		list( $amount, $unit, $upperBound, $lowerBound ) = unserialize( $data );
+		$amount = DecimalValue::newFromArray( $amount );
+		$upperBound = DecimalValue::newFromArray( $upperBound );
+		$lowerBound = DecimalValue::newFromArray( $lowerBound );
```
thiemowmde: I wonder what happened here? Did the serialization format changed?

src/DataValues/QuantityValue.php
```diff
@@ -473,14 +322,19 @@ public function getArrayValue() {
...
+	 * @since 0.7
 	 *
 	 * @param mixed $data
 	 *
-	 * @return self
+	 * @return QuantityValue
 	 * @throws IllegalValueException
 	 */
 	public static function newFromArray( $data ) {
+		if ( !isset( $data['upperBound'] ) && isset( $data['lowerBound'] ) ) {
```
thiemowmde: There is an `!` missing. A test should catch all possible cases.

src/DataValues/UnboundedQuantityValue.php
```diff
@@ -0,0 +1,334 @@
...
+ * The amount is stored as a @see DecimalValue object.
+ *
+ * @see QuantityValue for quantities with a known uncertainty interval.
+ * For simple numeric amounts use @see NumberValue.
+ *
+ * @note UnboundedQuantityValue and QuantityValue both use the value type ID "quantity".
+ * The fact that we use subclassing to model the bounded vs the unbounded case should be
+ * considered an implementation detail.
+ *
+ * @since 0.1
```
thiemowmde: 0.7

src/DataValues/UnboundedQuantityValue.php
```diff
@@ -0,0 +1,334 @@
...
+	public static function newFromNumber( $number, $unit = '1' ) {
+		$number = self::asDecimalValue( 'amount', $number );
+
+		return new self( $number, $unit );
+	}
+
+	/**
+	 * @see newFromNumber
+	 *
+	 * @deprecated since 0.1, use newFromNumber instead
```
thiemowmde: Introducing a deprecated method into a new class is odd.

src/DataValues/UnboundedQuantityValue.php
```diff
@@ -0,0 +1,334 @@
...
+	/**
+	 * @see Serializable::unserialize
+	 *
+	 * @since 0.1
+	 *
+	 * @param string $data
+	 */
+	public function unserialize( $data ) {
+		list( $amount, $unit ) = unserialize( $data );
+		$amount = DecimalValue::newFromArray( $amount );
```
thiemowmde: Again, I wonder why you introduced this. Probably for type safety and early exceptions, but the `__constructor` and `asDecimalValue` already do this.

src/DataValues/UnboundedQuantityValue.php
```diff
@@ -0,0 +1,334 @@
...
+	 * @return float
+	 */
+	public function getSortKey() {
+		return $this->amount->getValueFloat();
+	}
+
+	/**
+	 * Returns the quantity object.
+	 *
+*@see DataValue::getValue
```
thiemowmde: Indention.

src/DataValues/UnboundedQuantityValue.php
```diff
@@ -0,0 +1,334 @@
...
+
+	/**
+	 * @see Serializable::serialize
+	 *
+	 * @since 0.1
+	 *
+	 * @return string
+	 */
+	public function serialize() {
+		return serialize( array_values( $this->getArrayValue() ) );
```
thiemowmde: I just realized that this **is** a change to the serialization format. Side effects?

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -2,9 +2,10 @@
 
 namespace ValueFormatters;
 
+use DataValues\QuantityValue;
```
thiemowmde: Odd non-alphabetic reordering.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -26,6 +27,24 @@ class QuantityFormatter extends ValueFormatterBase {
...
+	 * Value for the OPT_SHOW_UNCERTAINTY_MARGIN indicating that the uncertainty margin
+	 * should be shown if we are displaying a QuantityValue.
+	 */
+	const SHOW_UNCERTAINTY_MARGIN_IF_KNOWN = 'if known';
+
+	/**
+	 * Value for the OPT_SHOW_UNCERTAINTY_MARGIN indicating that the uncertainty margin
+	 * should be shown if we are displaying a non-exact QuantityValue.
+	 */
+	const SHOW_UNCERTAINTY_MARGIN_IF_NOT_ZERO = 'if not zero';
```
thiemowmde: 1. I believe this must be a separate patch.
2. Using `…&showQuantityUncertaintyMargin=if%20not%20zero` in an URL is odd. Please avoid spaces.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -150,43 +172,50 @@ protected function formatQuantityValue( QuantityValue $quantity ) {
...
 	 *
 	 * @return int
 	 */
-	private function getRoundingExponent( QuantityValue $quantity ) {
-		if ( $this->options->getOption( self::OPT_APPLY_ROUNDING ) === true ) {
-			// round to the order of uncertainty
+	private function getRoundingExponent( UnboundedQuantityValue $quantity, $roundingMode ) {
+		if ( is_int( $roundingMode ) ) {
+			// round to the given exponent
+			return $roundingMode;
```
thiemowmde: Support for integers in a boolean option that's called "apply rounding"? Sorry, not convinced. Especially because the numbers are base-10 exponents, and conflict with what lower and upper bound are meant to do.

Again, this looks like a new feature that **must** be in it's own pull request.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -197,16 +226,22 @@ private function getRoundingExponent( QuantityValue $quantity ) {
...
 
-			if ( !$roundedMargin->isZero() ) {
-				return $this->decimalFormatter->format( $roundedMargin );
-			}
+		if ( $marginModel === self::SHOW_UNCERTAINTY_MARGIN_NEVER ) {
+			return null;
+		}
+
+		// TODO: never round to 0! See bug #56892
+		$roundedMargin = $this->decimalMath->roundToExponent( $margin, $roundingExponent );
```
thiemowmde: Applying the same (guessed) rounding to both the base number and the margin never was correct.

tests/DataValues/QuantityValueTest.php
```diff
@@ -290,17 +261,17 @@ public function transformProvider() {
 		};
 
 		return array(
-			 0 => array( QuantityValue::newFromNumber( '+10',   '1', '+11',  '+9' ),   $identity, QuantityValue::newFromNumber(   '+10',    '?',   '+11',    '+9' ) ),
```
thiemowmde: These whitespace changes make it hard to review this.

thiemowmde:
In general, I would love to start over with a lot of the implementation details in here, but can't as long as we are stuck with all this and not allowed to remove even //unused// code before we continue. As long as there is any code in our code base that turns 1.5±1 into 2±1 and 2.9±1.9 into 3±2 we are doing it **wrong**.

composer.json
brightbyte: I added it to this patch so I can uses phpunit to test this patch :)

src/DataValues/QuantityValue.php
brightbyte: THis is a sideeffect of renaming this back-and-forth. In the first version of this change, this code was in the new BoundedQuantityValue class. I'll fix all the since tags.

src/DataValues/QuantityValue.php
brightbyte: ok

src/ValueFormatters/QuantityFormatter.php
brightbyte: as discussed

src/ValueFormatters/QuantityFormatter.php
brightbyte: This is not new, we have been supporting int values for this option for years.

src/ValueFormatters/QuantityFormatter.php
brightbyte: Giving a margin of uncertainty with more precision than the value itself is also never correct.
But I agree that we should just show this as entered, correct or not.

brightbyte:
Updated the patch as discussed. A few notes:

* Removed unused getSignificantDigits method. Note that transform() will be used by https://gerrit.wikimedia.org/r/#/c/296962/
* Postponed the changing rounding behavior according to T95425 resp T105623 until we have QuantityMath
* New options in the formatter stay in, since we need new behavior for unbound quantities anyway.

brightbyte:
*sigh* after the rebase, lots of tests fail. no time to look into that now.

composer.json
JeroenDeDauw: Clearly it was possible to test the code before you made this change?

src/ValueFormatters/QuantityFormatter.php
thiemowmde: The code I'm complaining about is green, marked as new. I wonder what made it necessary to rearrange it. Did you re-introduced the formerly removed boolean options?

src/DataValues/QuantityValue.php
```diff
@@ -473,14 +342,22 @@ public function getArrayValue() {
...
 	 *
+	 * @note: if the upperBound or lowerBound field is missing from $data,
+	 * this returns an UnboundedQuantityValue, not a QuantityValue!
+	 *
 	 * @since 0.1
 	 *
 	 * @param mixed $data
 	 *
-	 * @return self
+	 * @return UnboundedQuantityValue
```
thiemowmde: Callers should be aware that this can return `UnboundedQuantityValue|self`.

src/DataValues/QuantityValue.php
```diff
@@ -473,14 +342,22 @@ public function getArrayValue() {
...
 	 * @throws IllegalValueException
 	 */
 	public static function newFromArray( $data ) {
+		if ( !isset( $data['upperBound'] ) || !isset( $data['lowerBound'] ) ) {
```
thiemowmde: Must be `&&`. As far as I remember I said this before: Please add tests for at least one possible combination to cover this.

src/DataValues/UnboundedQuantityValue.php
```diff
@@ -0,0 +1,323 @@
...
+ * The amount is stored as a @see DecimalValue object.
+ *
+ * @see QuantityValue for quantities with a known uncertainty interval.
+ * For simple numeric amounts use @see NumberValue.
+ *
+ * @note UnboundedQuantityValue and QuantityValue both use the value type ID "quantity".
+ * The fact that we use subclassing to model the bounded vs the unbounded case should be
+ * considered an implementation detail.
+ *
+ * @since 0.7
```
thiemowmde: 0.7 is already released.

thiemowmde:
Would be easier to have the parser changes in a separate patch. I probably have not said that clear enough when we discussed this.

thiemowmde:
+706 −294. :-( Can we please, please split this patch?

1. Add the new base class. Done.
2. Change the formatter to accept the base class.
3. Change the parser to return the base class.

composer.json
brightbyte: Yes, if you happened to have the correct version of phpunit installed. If you didn't, you had to guess which version would work.

thiemowmde:
To what does "Removed unused getSignificantDigits method" refer to? There is no such method.

src/DataValues/QuantityValue.php
brightbyte: Uh, I removed this, why is it back?? Damn rebase hell :(

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -26,6 +27,24 @@ class QuantityFormatter extends ValueFormatterBase {
...
+	 * Value for the OPT_SHOW_UNCERTAINTY_MARGIN indicating that the uncertainty margin
+	 * should be shown if we are displaying a QuantityValue.
+	 */
+	const SHOW_UNCERTAINTY_MARGIN_IF_KNOWN = 'if-known';
+
+	/**
+	 * Value for the OPT_SHOW_UNCERTAINTY_MARGIN indicating that the uncertainty margin
+	 * should be shown if we are displaying a non-exact QuantityValue.
+	 */
+	const SHOW_UNCERTAINTY_MARGIN_IF_NOT_ZERO = 'if-not-zero';
```
thiemowmde: This means we will mix camel case with dash style: `showQuantityUncertaintyMargin=if-not-zero`. I wonder if we should have a style guide for this.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -178,23 +219,25 @@ protected function formatNumber( QuantityValue $quantity ) {
...
-			return null;
+	private function getRoundingExponent( UnboundedQuantityValue $quantity, $roundingMode ) {
+		if ( is_int( $roundingMode ) ) {
+			// round to the given exponent
+			return $roundingMode;
+		} elseif ( $roundingMode && ( $quantity instanceof  QuantityValue ) ) {
+			// round to the order of uncertainty (QuantityValue only)
+			return $quantity->getOrderOfUncertainty();
 		} else {
-			return (int)$this->options->getOption( self::OPT_APPLY_ROUNDING );
```
thiemowmde: This looks like an unrelated breaking change. Before the code accepted numeric strings. Now it seems it ignores the option if e.g. `"2"` is provided as a string.

src/DataValues/QuantityValue.php
brightbyte: With &&, if only one of the fields is missing, we'll die with a fatal error. Is that desirable?

src/ValueParsers/QuantityParser.php
```diff
@@ -2,10 +2,11 @@
 
 namespace ValueParsers;
 
+use DataValues\QuantityValue;
```
thiemowmde: Why did this moved?

src/DataValues/QuantityValue.php
brightbyte: Callers should only care that the result is guaranteed to be an UnboundedQuantityValue, and not guaranteed to be a QuantityValue.

src/ValueFormatters/QuantityFormatter.php
brightbyte: I kind of like camel case for names and dashes for values... maybe I just have a strange taste.

src/DataValues/QuantityValue.php
thiemowmde: The formatter **must** know this can return a bound value, otherwise we can not display the uncertainty any more.

brightbyte: It does know that. Anything that can return an UnboundedQuantityValue can also return a QuantityValue.

thiemowmde: The documentation should say that then.

src/DataValues/UnboundedQuantityValue.php
```diff
@@ -0,0 +1,323 @@
...
+
+		if ( $number === null ) {
+			if ( $default === null ) {
+				throw new InvalidArgumentException( '$' . $name . ' must not be null' );
+			}
+
+			$number = $default;
+		}
+
+		if ( $number instanceof UnboundedQuantityValue ) {
```
thiemowmde: We decided this should not be done.

PR #67, opened 2016-07-06
Make public function to convert numeric string or value to DecimalValue.

smalyshev:
DecimalMath has it as private, but other code may need the same function.

### DIFF
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -58,7 +58,9 @@ class DecimalValue extends DataValueObject {
 		}
 
 		$value = trim( $value );
-
+		if ( $value !== '' && $value[0] !== '-' && $value[0] !== '+' ) {
+			$value = '+' . $value;
+		}
 		if ( strlen( $value ) > 127 ) {
 			throw new IllegalValueException( 'Value must be at most 127 characters long.' );
 		}
--- a/tests/DataValues/DecimalValueTest.php
+++ b/tests/DataValues/DecimalValueTest.php
@@ -42,6 +42,8 @@ class DecimalValueTest extends DataValueTest {
 		$argLists[] = array( '+0.0' );
 		$argLists[] = array( '+0.000' );
 		$argLists[] = array( '+1.0' . str_repeat( ' ', 124 ) );
+		$argLists[] = array( '4.2' );
+		$argLists[] = array( ' 4.2' );
 
 		return $argLists;
 	}
@@ -51,7 +53,6 @@ class DecimalValueTest extends DataValueTest {
 
 		$argLists[] = array( 'foo' );
 		$argLists[] = array( '' );
-		$argLists[] = array( '4.2' );
 		$argLists[] = array( '++4.2' );
 		$argLists[] = array( '--4.2' );
 		$argLists[] = array( '-+4.2' );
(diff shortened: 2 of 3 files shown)


src/DataValues/DecimalValue.php
```diff
@@ -353,6 +353,24 @@ public static function newFromArray( $data ) {
...
+	public static function makeDecimalValue( $number ) {
+
+		if ( is_string( $number ) && $number !== '' ) {
+			if ( $number[0] !== '-' && $number[0] !== '+' ) {
+				$number = '+' . $number;
+			}
+		}
+
+		return new static( $number );
+	}
```
JeroenDeDauw: Any reason this is not done in the constructor?

src/DataValues/DecimalValue.php
```diff
@@ -353,6 +353,24 @@ public static function newFromArray( $data ) {
 	}
 
 	/**
+	 * Create DecimalValue from regular numeric string or value.
```
JeroenDeDauw: Creates a DecimalValue from a regular numeric string or value.

src/DataValues/DecimalValue.php
smalyshev: I'd be ok with either way. The main thing is ability to accept string like "1234".

thiemowmde: The constructor already accepts ints and floats, therefore I do not see why it should not accept a simple string like "1234" with no leading plus sign. Please move to the constructor and add tests.

The nested `if`s are a bit weird in my opinion. Should be after the `trim` and more like `if ( ctype_digit( $number[0] ) )`.

smalyshev:
Moved to ctor and fixed the test.

tests/DataValues/DecimalValueTest.php
```diff
@@ -42,6 +42,7 @@ public function validConstructorArgumentsProvider() {
 		$argLists[] = array( '+0.0' );
 		$argLists[] = array( '+0.000' );
 		$argLists[] = array( '+1.0' . str_repeat( ' ', 124 ) );
+		$argLists[] = array( '4.2' );
```
JeroenDeDauw: `' 4.2'` should also be valid no?

Also need to verify the value gets interpreted correctly. If I'm not mistaken, you could change the production code to add a minus instead of a plus now, and the tests would still pass.

smalyshev: I didn't have a goal of making ' 4.2' (with space) valid, since no math function (like bcmath) would return such number.

src/DataValues/DecimalValue.php
```diff
@@ -58,7 +58,9 @@ public function __construct( $value ) {
 		}
 
 		$value = trim( $value );
-
+		if ( $value !== '' && $value[0] !== '-' && $value[0] !== '+' ) {
```
thiemowmde: I don't think this is the best implementation. It will add plus signs in front of strings that don't start with a number. It will, for example, produce `"+−1"` when the string starts with an actual minus sign. What was wrong with my suggestion?

smalyshev: I'm not sure I understand. The clause has specific check for $value[0] != '-' - could you explain how would it add + to -1?

thiemowmde: I added a few more test cases in #70 but could not find a way to outsmart the current implementation, so it is fine for now. It could, in theory, become a problem when we want to have support for [U+2212](https://en.wikipedia.org/wiki/Plus_and_minus_signs#Minus_sign).

PR #68, opened 2016-07-08
Fix obscure rounding in QuantityFormatter

thiemowmde:
This finally fixes https://phabricator.wikimedia.org/T95425.

Main changes in this patch:
* Never round the uncertainty margin. This never made sense, because the uncertainty margin is what *defines* the rounding. Rounding this number means it's changed based on information that's extracted from the same number. The number rounds itself.
* Do not round when the uncertainty margin is rendered. There is no need to do this. 1.5783+/-1 is not more unclear, confusing or wrong than 2+/-1, the contrary.
* These changes mean the default rounding option (true) does not have an effect any more when the margin option is also true. Integer rounding options are still possible. Disabling rounding would be wrong because this would also disable rounding when the margin is not rendered and rounding is crucial.
* Add much more test cases for the current behaviour of the code (some still with FIXMEs, but I did not wanted to make this patch bigger).

### DIFF
--- a/src/ValueFormatters/QuantityFormatter.php
+++ b/src/ValueFormatters/QuantityFormatter.php
@@ -158,10 +158,17 @@ class QuantityFormatter extends ValueFormatterBase {
 		$roundingExponent = $this->getRoundingExponent( $quantity );
 
 		$amount = $quantity->getAmount();
-		$roundedAmount = $this->decimalMath->roundToExponent( $amount, $roundingExponent );
-		$formatted = $this->decimalFormatter->format( $roundedAmount );
 
-		$margin = $this->formatMargin( $quantity->getUncertaintyMargin(), $roundingExponent );
+		if ( $roundingExponent === null ) {
+			$formatted = $this->formatMinimalDecimal( $amount );
+			$margin = $quantity->getUncertaintyMargin();
+			$margin = $margin->isZero() ? null : $this->formatMinimalDecimal( $margin );
+		} else {
+			$roundedAmount = $this->decimalMath->roundToExponent( $amount, $roundingExponent );
+			$formatted = $this->decimalFormatter->format( $roundedAmount );
+			$margin = $this->formatMargin( $quantity->getUncertaintyMargin(), $roundingExponent );
+		}
+
 		if ( $margin !== null ) {
 			// TODO: use localizable pattern for constructing the output.
 			$formatted .= '±' . $margin;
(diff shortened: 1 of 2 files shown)


thiemowmde:
Pinging @brightbyte, because this would, in my opinion, help to unstuck almost all other pull requests in this component.

brightbyte:
Just one comment: rounding needs to be applied after unit conversion even then the margin is rendered. One way to do this is to apply the rounding during the conversion, not when rendering. However, in that case, the rounding also needs to apply to the margin. Otherwise 13+/-2ft becomes 3.9624+/-0.6096m instead of the expected 4+/-0.6m.

thiemowmde:
I tried to say this before (not sure where): Apply rounding during conversion, with an algorithm that makes sure the inverse conversion back to the original unit results in the same value. Such an algorithm must know the original value and both units, therefore it must be applied during conversion.

This may mean that 13+/-2ft becomes 3.96+/-0.6m internally, but is rendered as 4m when rendered without the margin.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -158,10 +158,17 @@ protected function formatNumber( QuantityValue $quantity ) {
 		$roundingExponent = $this->getRoundingExponent( $quantity );
 
 		$amount = $quantity->getAmount();
-		$roundedAmount = $this->decimalMath->roundToExponent( $amount, $roundingExponent );
-		$formatted = $this->decimalFormatter->format( $roundedAmount );
 
-		$margin = $this->formatMargin( $quantity->getUncertaintyMargin(), $roundingExponent );
+		if ( $roundingExponent === null ) {
```
brightbyte: Keep in mind that UnboundedQuantityValues should only be rounded if that is explicitly requested by the user.

tests/ValueFormatters/QuantityFormatterTest.php
```diff
@@ -97,23 +97,65 @@ public function validProvider() {
...
 			'+3.025/xr' => array( QuantityValue::newFromNumber( '+3.025', '1', '+3.02744', '+3.0211' ), '3.03', $exactRounding ), // TODO: never round to 0! See bug #56892
 
 			'+3.125/nr' => array( QuantityValue::newFromNumber( '+3.125', '1', '+3.2', '+3.0' ), '3.125±0.125', $noRounding ),
 			'+3.125/xr' => array( QuantityValue::newFromNumber( '+3.125', '1', '+3.2', '+3.0' ), '3.13±0.13', $exactRounding ),
 
 			'+3.125/fs' => array( QuantityValue::newFromNumber( '+3.125', '1', '+3.2', '+3.0' ), '+3.13', $forceSign ),
+
+			// Rounding with a fixed +/-1 margin
+			array( QuantityValue::newFromNumber( '+1.44', '1', '+2.44', '+0.44' ), '1', $noMargin ),
+			// FIXME: Rounding this up is just wrong.
```
brightbyte: Good find! Looks like i managed to confuse myself there. Filed https://phabricator.wikimedia.org/T139748

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -158,10 +158,17 @@ protected function formatNumber( QuantityValue $quantity ) {
...
+			$formatted = $this->formatMinimalDecimal( $amount );
+			$margin = $quantity->getUncertaintyMargin();
+			$margin = $margin->isZero() ? null : $this->formatMinimalDecimal( $margin );
```
brightbyte: This still hides the uncertainty margin when it is zero. We agreed to change this, but perhaps not in this patch.

thiemowmde: Such a change absolutely **must** be it's own patch.

src/ValueFormatters/QuantityFormatter.php
thiemowmde: Unrelated. There is no UnboundedQuantityValues yet.

PR #69, opened 2016-07-09
Fix decimal string rounding algoithm.

brightbyte:
There's nothing harder than simple arithmetics…

[Bug: T139748](http://phabricator.wikimedia.org/T139748)

### DIFF
--- a/src/DataValues/DecimalMath.php
+++ b/src/DataValues/DecimalMath.php
@@ -234,108 +234,58 @@ class DecimalMath {
 			return '+0';
 		}
 
-		// whether the last character is already part of the integer part of the decimal value
-		$inIntPart = ( strpos( $value, '.' ) === false );
-		$rounded = '';
+		$len = strlen( $value );
 
-		// Iterate over characters from right to left and build the result back to front.
-		for ( $i = strlen( $value ) -1; $i > 0 && $i > $significantDigits; $i-- ) {
-			list( $value, $i, $inIntPart, $next ) = $this->roundNextDigit( $value, $i, $inIntPart );
-			$rounded = $next . $rounded;
+		// keeping all digits means no rounding
+		if ( $significantDigits >= ( $len -1 ) ) {
+			return $value;
 		}
 
-		// just keep the remainder of the value as is (this includes the sign)
-		$rounded = substr( $value, 0, $i +1 ) . $rounded;
+		// whether the last character is already part of the integer part of the decimal value
+		$i = min( $significantDigits +1, $len ); // account for the sign
+		$ch = $i < $len ? $value[$i] : '0';
 
-		if ( strlen( $rounded ) < $significantDigits + 1 ) {
-			if ( $inIntPart ) {
-				$rounded .= '.';
+		if ( $ch === '.' ) {
+			// NOTE: we expect the input to be well formed, so it cannot end with a '.'
+			$i++;
+			$ch = $i < $len ? $value[$i] : '0';
+		}
+
+		// split in significant and insignificant part
+		$rounded = substr( $value, 0, $i );
+
+		if ( strpos( $rounded, '.' ) === false ) {
+			$suffix = substr( $value, $i );
+
+			// strip insignificant digits after the decimal point
+			$ppos = strpos( $suffix, '.' );
+			if ( $ppos !== false ) {
+				$suffix = substr( $suffix, 0, $ppos );
 			}
 
-			$rounded = str_pad( $rounded, $significantDigits+1, '0', STR_PAD_RIGHT );
+			// zero out insignificant digits
+			$suffix = strtr( $suffix, '123456789', '000000000' );
+		} else {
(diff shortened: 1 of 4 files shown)


src/DataValues/DecimalMath.php
```diff
@@ -252,115 +252,62 @@ private function roundDigits( $value, $significantDigits ) {
 			return '+0';
 		}
 
+		$len = strlen( $value );
+
+		// keeping all digits means no rounding
+		if ( $significantDigits >= ( $len -1 ) ) {
```
thiemowmde: I still don't fully understand what an integer number called "significant digits" actually means. Starting from where? Counting what?

thiemowmde: How many "significant digits" does `"1230"` have? Should be 3, right? Which is equal to the string length minus one.

What if the passed `$value` is `"1234"` but `$significantDigits` is 3?

Why minus one?

thiemowmde: In other words: Is `$value` guaranteed to have a sign character and a decimal point? If not, the length calculation must do something about this, because `"1234"`, `"1.23"` and `"-1.2"` all have the same string length, but three different significant digits.

src/DataValues/DecimalMath.php
```diff
@@ -252,115 +252,62 @@ private function roundDigits( $value, $significantDigits ) {
...
+			if ( $ppos !== false ) {
+				$suffix = substr( $suffix, 0, $ppos );
+			}
 
-		if ( $ch === '.' ) {
-			// just transition from the fractional to the integer part
-			$inIntPart = true;
-			$nextCharacter = '.';
+			// zero out insignificant digits
+			$suffix = strtr( $suffix, '123456789', '000000000' );
```
thiemowmde: This should be pointless because there is nothing after these zeros that does make a difference. Simply truncate the string.

src/DataValues/DecimalMath.php
```diff
@@ -252,115 +252,62 @@ private function roundDigits( $value, $significantDigits ) {
...
-			}
+			// decimal point is in $rounded, so $suffix is insignificant
+			$suffix = '';
+		}
 
-			if ( ord( $ch ) >= ord( '5' ) ) {
-				// when stripping a character >= 5, bump up the next digit to the left.
-				list( $value, $i, $inIntPart ) = $this->bumpDigitsForRounding( $value, $i, $inIntPart );
-			}
+		if ( ord( $ch ) >= ord( '5' ) ) {
```
thiemowmde: The two `ord` functions are not needed. Either compare these as strings with `strcmp` or rely on PHP's magic casting both to int.

src/DataValues/DecimalValue.php
```diff
@@ -63,7 +63,7 @@ public function __construct( $value ) {
 			throw new IllegalValueException( 'Value must be at most 127 characters long.' );
 		}
 		if ( !preg_match( self::QUANTITY_VALUE_PATTERN, $value ) ) {
-			throw new IllegalValueException( 'Value must match the pattern for decimal values.' );
+			throw new IllegalValueException( 'Value must match the pattern for decimal values. Malformed: ' . $value );
```
thiemowmde: Weird technical wording. Suggestion: `"\"$value\" is not a well formed decimal value"`.

src/DataValues/DecimalMath.php
brightbyte: $significantDigits is documented in roundToDigit(): $significantDigits The number of digits to retain, counting the decimal point, but not counting the leading sign.

Counting the decimal point makes it easier to use $significantDigits as a string offset. But it's still bad, we should change it. But I didn't want to mess with it in this change. 

The -1 is needed because the first character is the sign. It's "+1234", not "1234".

So: the sign is guaranteed to be there, and $significantDigits includes the decimal point (even though that is silly). That's why the offset calculation works.

src/DataValues/DecimalMath.php
brightbyte: if $value is 12377 and has 3 significant digits, i would truncate it to 123, then. The 77 needs to be replaced with 00 instead. There are of course various ways to do that, but this one seemed fast & simple.

src/DataValues/DecimalMath.php
brightbyte: Comparing directly to 5 feels dirty. But comparing directly to '5' would be fine, I guess.

src/DataValues/DecimalValue.php
brightbyte: Yea, i just slapped that on while debugging.

brightbyte:
hm, scrutenizer failure looks like a bug on their side

thiemowmde:
I find specific implementation details problematic, as discussed. But this does not matter if the tests cover everything.

I rebased this to make sure it is tested against DataValues Interfaces 0.2.2, which fixed a bug relevant for this. +2 when Travis turns green.

PR #70, opened 2016-07-11
Fix integer overflow in DecimalValue

thiemowmde:
This applies to int32 as well as int64 PHP versions.


brightbyte:
What exactly caused the overflow?

thiemowmde:
The line I fixed? `$number === floor( $number )` is not a proper "can be converted to integer" check.

brightbyte:
Ah, it's the line *after* the one you changed that actually overflows. I was confused about floor() having a problem with big numbers...

PR #71, opened 2016-07-11
Add IEEE edge cases to QuantityFormatter test

thiemowmde:
I think the formatter should do something about such cases when the rounding option is set to true. Not exactly sure what and how. What it did before #68 was not acceptable, in my opinion. For now let's at least add test cases.


brightbyte:
If bcmath is installed, we should not have any trouble with floating point errors. Still, nice to have tests.

brightbyte:
Uh, ther mergability test got stuck. How do I poke it?

thiemowmde:
What is a "mergability test"?

brightbyte:
the github thingy that sais "we are checking whether we can merge this" or something

PR #74, opened 2016-07-11
Remove deprecated newFromDecimal method

thiemowmde:
Split from #66 to make it ~~easier~~ possible to review.

PR #75, opened 2016-07-11
Drop dead code from quantity classes

thiemowmde:
Split from #66 to make it ~~easier~~ possible to review.


thiemowmde:
Pinging @brightbyte for #75, #74, #73 and possibly #72.

PR #76, opened 2016-07-12
Only add UnboundedQuantityValue

thiemowmde:
This is split from #66 with no change to the implementation:
* Both files are almost 100% copies from the existing files, except that all "bounds" stuff is removed.
* I removed all `@since` tags except for the one on top.
* I compared this with the current state of the QuantityValue class and applied all minor changes we did (e.g. in #72).


brightbyte:
*sigh* needs rebase :/

thiemowmde:
Rebased. I had an other look and I am still very happy with this pull request. I looked at the two tests. Now that there is a base class it could be that the subclass is testing code paths that are already fully covered by the base class' test. Turns out: no. All test cases are different in the two tests, none is redundant.

PR #77, opened 2016-07-12
Add tests for QuantityValue::newFromArray

thiemowmde:
Split from #66. Pinging @brightbyte again.

### DIFF
--- a/tests/DataValues/QuantityValueTest.php
+++ b/tests/DataValues/QuantityValueTest.php
@@ -186,6 +186,12 @@ class QuantityValueTest extends DataValueTest {
 					'upperBound' => '+2.5',
 				)
 			),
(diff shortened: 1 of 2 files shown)


brightbyte:
Will need to change quite a bit once UnboundedQuantityValue is introduced.

thiemowmde:
Not sure what you mean. This is an untouched copy from the code currently in #66, except for one removed test case.

brightbyte:
I just merged UnboundedQuantityValue. Can you update this to add all the newFromArray tests now, for both classes?

thiemowmde:
This is already the case. As said, only a single test case from #66 is missing at the moment, because QuantityValue is still in a state where it behaves exactly like before. I did added that test case for now, see the 2nd commit.

PR #79, opened 2016-07-15
Allow formatting UnboundedQuantityValues

thiemowmde:
This is split from #66. I pulled all new test cases from #66, but not the implementation. Instead I tried to find the most minimal change that makes all tests succeed. Plus I added more test cases, especially for the edge case "do not add additional zeros when rounding to a specific number of decimal places". This got fixed by #69.

[Bug: T139748](https://phabricator.wikimedia.org/T139748)

### DIFF
--- a/src/ValueFormatters/QuantityFormatter.php
+++ b/src/ValueFormatters/QuantityFormatter.php
@@ -5,6 +5,7 @@ namespace ValueFormatters;
 use DataValues\DecimalMath;
 use DataValues\DecimalValue;
 use DataValues\QuantityValue;
+use DataValues\UnboundedQuantityValue;
 use InvalidArgumentException;
 
 /**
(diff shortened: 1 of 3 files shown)


thiemowmde:
Rebased. Pinging @brightbyte.

src/ValueFormatters/QuantityFormatter.php
```diff
@@ -148,6 +151,22 @@ protected function formatQuantityValue( QuantityValue $quantity ) {
...
+	 *
+	 * @return string
+	 */
+	protected function formatUnboundedQuantityValue( UnboundedQuantityValue $quantity ) {
+		$amount = $quantity->getAmount();
+		$roundingExponent = $this->options->getOption( self::OPT_APPLY_ROUNDING );
+
+		if ( !is_bool( $roundingExponent ) ) {
+			$amount = $this->decimalMath->roundToExponent( $amount, (int)$roundingExponent );
+		}
```
brightbyte: If $roundingExponent === true, we should guess the precision, and then apply rounding.

brightbyte: It's a bit odd that we don't apply rounding even if OPT_APPLY_ROUNDING is true...

thiemowmde: But if all you have is the string "1.499999999999999", no bounds and an option set to `true`, how do you round?

brightbyte: Yea, the idea is that the automatic precision guessing would guess a precision that would not change the number when rounding is applied. So it's equivalent to no rounding. Still, it feels odd.

PR #83, opened 2016-07-21
When guessing precision from digits, use +/-0.5 as the margin, not +/-1

brightbyte:
That is, 12 is interpreted as 12+/-0.5 = [11.5, 12.5),
instead of 12+/-1 = [11,13). This is consistent with rounding:
all numbers in the uncertainty interval will round to the original
number.

Bug: T105623

### DIFF
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -327,8 +327,9 @@ class DecimalValue extends DataValueObject {
 	 *
 	 * @return DecimalValue
 	 */
-	public function trim() {
-		$trimmed = preg_replace( '/(\.[\d]+?)0+$/', '\1', $this->value );
+	public function getTrimmed() {
+		$trimmed = preg_replace( '/(\.\d+?)0+$/', '$1', $this->value );
+		$trimmed = preg_replace( '/(?<=\d)\.0*$/', '', $trimmed );
 
 		if ( $trimmed !== $this->value ) {
 			return new DecimalValue( $trimmed );
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -321,6 +321,23 @@ class DecimalValue extends DataValueObject {
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/DataValues/DecimalValue.php
```diff
@@ -321,6 +321,23 @@ public function getFractionalPart() {
...
 
 	/**
+	 * Returns a DecimalValue with the same digits as this one, but with any trailing zeros
+	 * after the decimal point removed. If there are no trailing zeros after the decimal
+	 * point, this method will return $this.
+	 *
+	 * @return DecimalValue
+	 */
+	public function trim() {
+		$trimmed = preg_replace( '/(\.[\d]+?)0+$/', '\1', $this->value );
```
thiemowmde: `[\d]` is the same as `\d`.

I would also trim `1.0` to `1`.

src/DataValues/DecimalValue.php
```diff
@@ -321,6 +321,23 @@ public function getFractionalPart() {
 	}
 
 	/**
+	 * Returns a DecimalValue with the same digits as this one, but with any trailing zeros
+	 * after the decimal point removed. If there are no trailing zeros after the decimal
+	 * point, this method will return $this.
+	 *
+	 * @return DecimalValue
+	 */
+	public function trim() {
```
thiemowmde: This sounds like it modifies the original value, but instead it returns a new one. I think it should say that.

thiemowmde:
:+1: No objections. Mostly nitpicks and naming suggestions. These math codes always look scary, but are fine obviously.

brightbyte:
The scrutinizer failure seems unrelated. Seems like it doesn't work with PHP < 5.5, but is still trying to use 5.3?

brightbyte:
ping @thiemowmde: can this go in now?

thiemowmde:
Whoever wants to merge this, feel free. I don't have time since there is a dozen other equally and more important things on my table.

PR #85, opened 2016-08-02
Fix incomplete (Unbounded)QuantityValue::newFromArray deserializers

thiemowmde:
I consider this a "fix" for the following reason:
* The parser returns either an unbounded or a bounded quantity value.
* The formatter accepts both.
* Both do have a `getArrayValue` method, which is what defines the serialization. The serializers don't know anything about the two classes.
* We are considering the fact that we split the code into two classes an implementation detail.
* Deserialization is done in `DataValueDeserializer`, which calls `newFromArray`. This should still be possible without requiring an "IF upper and lower bound are present THEN deserialize bounded quantity value ELSE deserialize unbounded quantity value". This knowledge should be inside of this component (because it is an implementation detail). Not in Wikibase.git.

### DIFF
--- a/src/DataValues/QuantityValue.php
+++ b/src/DataValues/QuantityValue.php
@@ -292,26 +292,4 @@ class QuantityValue extends UnboundedQuantityValue {
 		);
 	}
 
-	/**
-	 * Constructs a new instance of the DataValue from the provided data.
-	 * This can round-trip with @see getArrayValue
-	 *
-	 * @since 0.1
-	 *
-	 * @param mixed $data
-	 *
-	 * @return self
-	 * @throws IllegalValueException
-	 */
-	public static function newFromArray( $data ) {
-		self::requireArrayFields( $data, array( 'amount', 'unit', 'upperBound', 'lowerBound' ) );
-
-		return new static(
-			DecimalValue::newFromArray( $data['amount'] ),
-			$data['unit'],
-			DecimalValue::newFromArray( $data['upperBound'] ),
-			DecimalValue::newFromArray( $data['lowerBound'] )
-		);
-	}
-
 }
--- a/src/DataValues/UnboundedQuantityValue.php
+++ b/src/DataValues/UnboundedQuantityValue.php
@@ -247,21 +247,33 @@ class UnboundedQuantityValue extends DataValueObject {
 	}
 
 	/**
-	 * Constructs a new instance of the DataValue from the provided data.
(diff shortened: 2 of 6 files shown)


thiemowmde:
Pinging @brightbyte to have a look at this.

src/DataValues/QuantityValue.php
```diff
@@ -298,15 +298,19 @@ public function getArrayValue() {
...
 	 *
-	 * @return self
+	 * @return UnboundedQuantityValue|self
 	 * @throws IllegalValueException
 	 */
-	public static function newFromArray( $data ) {
+	public static function newFromArray( array $data ) {
+		if ( !isset( $data['upperBound'] ) && !isset( $data['lowerBound'] ) ) {
+			return parent::newFromArray( $data );
+		}
```
brightbyte: This is quite unexpected, please add a warning or at least an explanation to the documentation.

Perhaps we should consider having a separate function for this - or even a QuantityValueFactory class.

src/DataValues/UnboundedQuantityValue.php
```diff
@@ -250,15 +250,19 @@ public function getArrayValue() {
...
 	 *
-	 * @return self
+	 * @return self|QuantityValue
 	 * @throws IllegalValueException
 	 */
-	public static function newFromArray( $data ) {
+	public static function newFromArray( array $data ) {
+		if ( isset( $data['upperBound'] ) || isset( $data['lowerBound'] ) ) {
+			return QuantityValue::newFromArray( $data );
+		}
```
brightbyte: This is somewhat unexpected, please add an explanation to the documentation.

brightbyte:
+1 from me, +2 with better documentation.

src/DataValues/QuantityValue.php
thiemowmde: Not sure what you expect whats not already in `@return UnboundedQuantityValue|self`?

brightbyte: Technically, @return UnboundedQuantityValue|self is complete, but it's easy to miss. Since we are doing something quite unexpected (returning an incompatible object from a pseudo-constructor), a @warning seems in order.

thiemowmde:
Much, much simpler now. The fact that the implementation is split into two classes is considered an implementation detail. It's intended that the classes know each other, in both directions. There should only be one of these static `newFromArray` constructors.

PR #95, opened 2017-04-19
Update PHPCS and PHPUnit dependencies

thiemowmde:
* Update PHPCS requirement to use the same version as Wikibase.git.
* Update PHPCS rule set to be closer to the one in Wikibase.git.
* Add PHPUnit as a dependency, as we are doing this everywhere now. This is "incomplete" in so far that CI is not using this yet. But this already helps a lot when running tests locally.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
 		"data-values/common": "~0.3.0|~0.2.0"
 	},
 	"require-dev": {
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
And coverage stuff not working either

thiemowmde:
That was not the motivation of this patch. I would be very thankful if you could either update this patch or upload a follow up.

PR #98, opened 2017-04-20
Minor fixups to CI and make it run PHPCS

thiemowmde:


JeroenDeDauw:
I took the liberty of amending and have travis call `composer test` instead of `ci` for now, since the phpcs checks apparently do not pass yet.

The sudo setting is not required. It tells travis that sudo is not needed during execution which allows them to use cheaper containers or something like that.

JeroenDeDauw:
When I run the test locally I get a big pile of errors: https://pastebin.com/ew8SCZiX

> php --version
PHP 7.0.15-1ubuntu4 (cli) (built: Feb 28 2017 21:33:59) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.15-1ubuntu4, Copyright (c) 1999-2017, by Zend Technologies

thiemowmde:
* I removed th e`sudo` thing.
* I changed Travis back to call `composer ci`, but changed `composer ci` to not run PHPCS for now. I will fix this in an other patch.
* The test failures are because your local system does not have the bcmath PHP extension installed. I will try to add this as a dependency to composer.json, or make the tests more flexible.

I believe we should merge this as it is now.

.travis.yml
```diff
@@ -9,8 +9,7 @@ php:
   - 7.1
   - hhvm
 
-before_script:
-  - composer install --prefer-source
+install: travis_retry composer install
```
manicki: This drops --prefer-source intentionally?
I don't mind, just noticed.

thiemowmde: Half of the Travis scripts we have contain this "--prefer-source" parameter, half of them don't. But I don't know why, so I went for the more simple, more obvious thing and removed it.

manicki:
This looks good. Commit message does say something else the change actually does in the end, though. Do we care? Or should I just shut up and merge?

thiemowmde:
Removed.

manicki:
?

Commit message says: Make CI use PHPUnit installed via Composer
This change does not really do it, as it already happens on master.

This inconsistency is what I meant in my commet?

thiemowmde:
I changed the title 43 minutes ago.

thiemowmde:
I edited the title of this pull request an hour ago. I never look at the commit headlines here on GitHub. Now I found what you meant and changed it.

manicki:
Yeah, I sometimes fail to look at PR titles :) git > github

PR #99, opened 2017-04-21
Fix all remaining PHPCS errors

thiemowmde:
As requested via #98.


thiemowmde:
Travis error is unrelated.

PR #100, opened 2017-04-21
Require BCMath PHP extension via Composer

thiemowmde:
The production code also works without the BCMath extension, but the tests don't. I believe it's impossible to test both the same time (how would you disable a PHP module in the middle of a PHPUnit test?). I believe the best solution is to straight away require the extension, at least in all test scenarios.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
 		"data-values/common": "~0.3.0|~0.2.0"
 	},
 	"require-dev": {
+		"ext-bcmath": "*",
 		"mediawiki/mediawiki-codesniffer": ">=0.4 <0.8",
 		"phpunit/phpunit": "~4.8"
 	},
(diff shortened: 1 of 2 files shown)


manicki:
+1

Random note: if this is about some tests, not all of them (which I would assume is the case based on "The production code also works without the BCMath extension") relevant test could possibly also be skipped on system without the extension installed, no?

thiemowmde:
This patch is a response to what @JeroenDeDauw wrote in https://github.com/DataValues/Number/pull/98#issuecomment-296118971 (also see https://pastebin.com/ew8SCZiX). Currently there is a single `$this->markTestSkipped( 'bcmath library not available' )` in `DecimalMathTest::testProductWithBC`. We could mark all affected tests the same way, but to me this feels scary (a lot of stuff is entirely untested then) and tedious because it's not possible to test all this properly without constantly uninstalling and reinstalling the BCMath extension.

I believe this issue became worse with the latest HHVM, PHP 7.0.x and 7.1.x versions. From what I have seen different test scenarios fail, depending on the PHP/HHVM version you are using. This means we either need to disable more and more test cases, whenever a new one fails. Or we require the library with the consistent behaviour we need.

manicki:
Yeah, got that.

PR #104, opened 2017-05-16
Fix HHVM tests failing on Travis

thiemowmde:
Everything on Travis is currently blowing up because of this, see https://github.com/travis-ci/travis-ci/issues/7712. I would like to rebase the patches that are waiting for a review here, and need this fix to make the tests succeed again.

This also adds PHP 7.1 to the test matrix in addition to PHP 7.0 and HHVM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_script:
   - composer install --prefer-source
 
 script:
-  - phpunit
+  - composer test
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


manicki:
Added a commit that makes travis use the PHP Unit binary installed by composer, not the "global" phpunit. This was causing failures on hhvm build.

Also, apparently I accidentally also turned PHPCS checks that haven't been run on travis previously (!)

thiemowmde:
Have you seen #98? I think this was doing pretty much the same. I rebased #98 now and would love to see this merged, too (if CI succeeds).

PR #106, opened 2017-06-23
Better documentation for the DataValue::newFromArray contract

thiemowmde:
#105 already fixed it, this is just making the contract much more clear, I hope.

[Bug: T168681](https://phabricator.wikimedia.org/T168681)

### DIFF
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -48,7 +48,7 @@ class DecimalValue extends DataValueObject {
 	 * @param string|int|float $value If given as a string, the value must match
 	 *  QUANTITY_VALUE_PATTERN. The leading plus sign is optional.
 	 *
-	 * @throws IllegalValueException
+	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $value ) {
 		if ( is_int( $value ) || is_float( $value ) ) {
@@ -111,8 +111,6 @@ class DecimalValue extends DataValueObject {
 
 	/**
 	 * Compares this DecimalValue to another DecimalValue.
-	 *
-	 * @since 0.1
 	 *
 	 * @param self $that
 	 *
@@ -227,8 +225,6 @@ class DecimalValue extends DataValueObject {
 	/**
 	 * Returns the sign of the amount (+ or -).
 	 *
-	 * @since 0.1
-	 *
 	 * @return string "+" or "-".
 	 */
 	public function getSign() {
@@ -288,8 +284,6 @@ class DecimalValue extends DataValueObject {
 	 * Returns the integer part of the value, that is, the part before the decimal point,
 	 * without the sign.
 	 *
-	 * @since 0.1
-	 *
 	 * @return string
 	 */
 	public function getIntegerPart() {
@@ -305,8 +299,6 @@ class DecimalValue extends DataValueObject {
 	/**
 	 * Returns the fractional part of the value, that is, the part after the decimal point,
 	 * if any.
-	 *
-	 * @since 0.1
 	 *
 	 * @return string
 	 */
@@ -342,8 +334,6 @@ class DecimalValue extends DataValueObject {
 	 * Returns the value held by this object, as a float.
 	 * Equivalent to floatval( $this->getvalue() ).
 	 *
-	 * @since 0.1
-	 *
 	 * @return float
 	 */
 	public function getValueFloat() {
@@ -360,13 +350,20 @@ class DecimalValue extends DataValueObject {
 	}
 
 	/**
-	 * Constructs a new instance of the DataValue from the provided data.
-	 * This can round-trip with @see getArrayValue
+	 * Constructs a new instance from the provided data. Required for @see DataValueDeserializer.
+	 * This is expected to round-trip with @see getArrayValue.
 	 *
-	 * @param string|int|float $data
+	 * @deprecated since 0.8.3. Static DataValue::newFromArray constructors like this are
+	 *  underspecified (not in the DataValue interface), and misleadingly named (should be named
+	 *  newFromArrayValue). Instead, use DataValue builder callbacks in @see DataValueDeserializer.
 	 *
+	 * @param mixed $data Warning! Even if this is expected to be a value as returned by
+	 *  @see getArrayValue, callers of this specific newFromArray implementation can not guarantee
+	 *  this. This is not guaranteed to be a string!
+	 *
+	 * @throws InvalidArgumentException if $data is not in the expected format. Subclasses of
+	 *  InvalidArgumentException are expected and properly handled by @see DataValueDeserializer.
 	 * @return self
-	 * @throws IllegalValueException
 	 */
 	public static function newFromArray( $data ) {
 		return new static( $data );
--- a/src/DataValues/UnboundedQuantityValue.php
+++ b/src/DataValues/UnboundedQuantityValue.php
@@ -64,7 +64,7 @@ class UnboundedQuantityValue extends DataValueObject {
 	 * @param string $unit A unit identifier. Must not be empty, use "1" for unit-less quantities.
 	 *
 	 * @return self
-	 * @throws IllegalValueException
+	 * @throws InvalidArgumentException
 	 */
 	public static function newFromNumber( $amount, $unit = '1' ) {
 		$amount = self::asDecimalValue( 'amount', $amount );
@@ -82,7 +82,6 @@ class UnboundedQuantityValue extends DataValueObject {
 	 * @param string|int|float|DecimalValue|null $number
 	 * @param DecimalValue|null $default
 	 *
-	 * @throws IllegalValueException
 	 * @throws InvalidArgumentException
 	 * @return DecimalValue
 	 */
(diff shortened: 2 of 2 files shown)


src/DataValues/DecimalValue.php
```diff
@@ -360,13 +350,20 @@ public function getArrayValue() {
...
 
 	/**
-	 * Constructs a new instance of the DataValue from the provided data.
-	 * This can round-trip with @see getArrayValue
+	 * Constructs a new instance from the provided data. Required for @see DataValueDeserializer.
+	 * This is expected to round-trip with @see getArrayValue.
+	 *
+	 * @deprecated since 0.8.3. Static DataValue::newFromArray constructors like this are
+	 *  underspecified (not in the DataValue interface), and misleadingly named (should be named
+	 *  newFromArrayValue). Instead, use DataValue builder callbacks in @see DataValueDeserializer.
```
manicki: hmm, I am not sure if this is framed the best way. DataValueDeserializer indeed makes use of those constructors. But why would this library bother about some DataValueDeserializer? And, probably not in this case, in theory, there could be other users of those constructors than DataValueDeserializer.
Also, I note that builder callback for "wikibase-entityid" in Wikibase does use EntityIdValue::newFromArray :)

I know it is good to provide alternatives when deprecating a part of public interface, but maybe simple "DO NOT USE" would be enough here?

thiemowmde: The only reason the newFromArray methods exist is because DataValueDeserializer needs them. DataValue::getArrayValue serializes, newFromArray deserializes. That was the idea.

I don't care much about other code. What's relevant and really, really crucial here is to know how this is going to be used in DataValueDeserializer. I tried to add the contract of "newFromArray" including this big comment to the DataValue interface, but I figured this would make the situation worse. We already have an implementation without a "newFromArray", and we should be free to remove them individually when we are sure they are not used any more. That's why it should not be in the interface, and that's why the deprecation must mention the alternative.

manicki: Sure, I understand that. But as I pointed out, using "builder callback" does not mean it would not call newFromArray (which it does now).
From the perspective of this component this depreciation means: do not use newFromArray, use regular constructor, or do I get it wrong?

thiemowmde: I want to get rid of the bad code in DataValueDeserializer that just assumes every DataValue implementation does have a newFromArray method with a specific contract, because this contract is not specified anywhere.

The deprecation is mostly because of the bad name. Builder callbacks are still free to use whatever constructor they want (the default one or a more convenient static one).

manicki: I understand that. I do not oppose to depreciation or anything. I am only mentioning that I found the depreciation notice a bit confusing (a single sentence there, in particular).

I don't know this all DataValues stuff well, so it would be the best if @JeroenDeDauw or @brightbyte chimed in here I guess. It simply seems a bit odd to me that this components mentions some other component which is not even its dependency etc.

thiemowmde: There is a dependency. DataValueDeserializer is where the contract for the newFromArray methods is specified. This is not a hard dependency in the sense of code usage, but a dependency in the sense of information flow. So yes, there is a circular dependency here. I want to document this, and mark it as deprecated everywhere, so we can get rid of it some day.

brightbyte: I haven't looked at the code again, just a general comment about @deprecated: the current deprecation notice gives a justification for the deprecation. That's usually not what the caller needs - they just need to know since when this is deprecated, and what to do instead. 

Information about technical debt should probably for formulated as a @todo, in an actionable way. I guess it's ok to give a reason there. But perhaps it's sufficient if the justification is in the commit message.

thiemowmde: I don't understand what you guys want. I find it really, really critical to mention the reasoning for the deprecation, because it is two-fold: The only (!) reason "newFromArray" exists in the first place is because of this one line in the DataValueDeserializer. There was no alternative for this a while ago, but is now. So why not mention this alternative? And the name is just bad, which caused the current pain in the first place. I was under the impression a method that is literally called "newFromArray" always gets an array. But this is wrong. Users should introduce a new one with a better name, if they really need one.

brightbyte: First off: I don't have strong feelings about this, I'm just giving my experience with deprecation messages. 

So: the deprecation message should of course mention the alternative - that's the most important bit! And for callers, that (plus the deprecation version) is the only information that is relevant.

The justification is relevant for people who work on improving the library and addressing tech debt. So it should be in a @todo. Why do you think it is critical to give the justification in the deprecation message?

(Again, for the record: I think this is a minor point. An interesting detail discussion about how to write perfect documentation. Shouldn't block a patch.)

manicki: I also don't say this is something more than a minor point. I just wanted to point out that this alternative suggested here is provided from the perspective of one particular use of this component.
As the same/very similar notice if now being suggested to be add to N other related components, I'd argue it is good to have an accurate documentation everywhere.
If others say the current notice is fine, I am not going to oppose. As I don't oppose here.

thiemowmde: > for callers, that […] is the only information that is relevant.

It should have become clear that I disagree.

> Why do you think it is critical to give the justification in the deprecation message?

Because as a consumer of this I want to know what's going on, and why I have to change code that was just fine for months, possibly years. As a consumer I need to understand the reasoning so I can follow it and properly apply it to my own code. I will not blindly follow an instruction when it does not make sense to me, especially if there is no simple replacement that can be applied with a regular expression, as it is the case here. If there really is a third-party user outside of DataValueDeserializer he should talk to us and we will introduce something with a better name than the current "newFromArray". A `@todo` can not do this. I don't see how a third-party user should become aware of a `@todo` hidden in some otherwise unchanged code.

> […] provided from the perspective of one particular use of this component.

Yes, because there is only one. Secondary users of these methods don't exist, as far as we know.

filbertkm: Thanks for deprecating this. It's confusing that some of these methods (e.g. StringValue) take something other than an array in this method.

maybe people should be advised just to "use the constructor builder callbacks (e.g. @see DataValueDeserializer) instead".

thiemowmde:
It's 7 hours since I uploaded this patch. Then we spend almost 2 hours discussing what I did, without getting anywhere. Then everybody stopped. This makes me angry and helpless. I spend the entire morning on [T168681](https://phabricator.wikimedia.org/T168681) and all related patches. I was hoping we could do the bugfix releases today. Really, in essence they do nothing but removing two `array` type hints. How complicated could that be? It seems it can, even to a point where everything just gets stuck. Why? What are you missing? What did I do wrong? Do you think this is not relevant any more? Did I wasted my time? Who will continue working on this, if what I did is not sufficient?

src/DataValues/DecimalValue.php
thiemowmde: @filbertkm, I don't understand. This is what it pretty much literally says: use builder callbacks instead.

manicki:
> Then we spend almost 2 hours discussing what I did, without getting anywhere. 

This is not really constructive comment, speaking about "getting anywhere".

> Then everybody stopped. 

So I made come comments. I consider them really minor, and I hope made myself clear. To those you more or less said "This does not matter". I am not blocking anything here, even if I disagreed. I said I'd prefer people more involved in the development of this library to say how they would like this depreciation communicated.
I could go ahead and merge this patch. Still the bug fix will not be deployed to wikidata.org before Monday any way. Therefore I felt it is OK to wait a bit more and possible have wise people peep in here.

src/DataValues/DecimalValue.php
JeroenDeDauw: Such much comments... read the first few

`DataValueDeserializer` was added to allow getting rid of these static construction methods. Apparently no deprecation tag was added to these specific classes. Doing so should not be an issue. Perhaps removing them altogether also is not? If migration to `DataValueDeserializer` has been completed, you can just update the code in `DataValueDeserializer` and ditch these methods. About time after 4(?) years.

I agree that generally mentioning a class from a component you do not even depend on is either weird or flat out bad. Since the responsibility for this serialization and deserilaization stuff was moved to another component, it does seem like the sensible thing to do here. And anyway, no need to spend lots of time on writing the best comment ever if you just get rid of it and the code it goes along with in the near future (which is what I hope and suggest you do).

thiemowmde: DataValueDeserializer exists since 2013. Not only was it the only caller of the DataValue::newFromArray methods. The call in DataValueDeserializer was actually the specification of how all DataValue::newFromArray methods must behave. Which is why I started calling this a "circular dependency" in a comment above. An alternative was added in 2016 with https://github.com/DataValues/Serialization/pull/14.

JeroenDeDauw: > Not only was it the only caller of the DataValue::newFromArray methods. The call in DataValueDeserializer was actually the specification of how all DataValue::newFromArray methods must behave.

Wrong. As I wrote, `newFromArray` predates `DataValueDeserializer`, the later being added to migrate away from the former.

> Which is why I started calling this a "circular dependency" in a comment above.

That is poor use of words IMO, since this most certainly is not a circular dependency. You can broaden the term sure, but if you go this far, it becomes so vague that it loses its utility.

> An alternative was added in 2016 with DataValues/Serialization#14.

Yeah cool. This is the kind of thing that was always meant to happen to finish the migration.

---

This is just talking about the past. I'd rather focus on how to proceed now. Anyway, going to ubsub here and leave this in your (collective) capable hands.

manicki:
Thanks @JeroenDeDauw for explaining. With this information it was clear to me this should be deprecated.

When this PR was opened, it has probably been the second time I ever looked in any of this data values code. That's why I wanted someone else to confirm.

Now with all the comments, and looking some more, I can see what is the intended direction here. But really, it does not strike as obvious that newFromArray should have been deprecated and that this DataValueDeserializer is the replacement for newFromArray, when both DataValueDeserializer and its famous callbacks happily use newFromArray.

src/DataValues/DecimalValue.php
thiemowmde: > Wrong.

It appears you did not read what I wrote, or just don't want to understand what I meant. I find this really frustrating. Again: since 2013 the detailed specification of how all newFromArray methods must behave was not in the DataValue interface but in DataValueDeserializer. How do you call this, if not a circular dependency?

thiemowmde:
@manicki, technically, there is nothing wrong with having static constructors. Each DataValue implementation is entirely free to have one or more.

The main problem we are talking about here is neither caused by the pure existence of the newFromArray methods, nor by the callbacks using them, but solely by the fact that DataValueDeserializer expects them to exist and behave in a very specific way that is not specified in the DataValue interface. If this would be the only problem with the newFromArray methods, it would not be necessary to deprecate them.

But there are more problems, as I already tried to explain above. The name is misleading as not all of them expect an array. In simple cases like StringValue the method does not do anything but calling the default constructor. Also the deprecation makes all the problems we are talking about here much, much more obvious. A `@todo` somewhere in DataValueDeserializer or a Phabricator ticket can't do this.

manicki:
hmm, then if in principle it would be OK to keep static newFromArray constructors (possible having them renamed), why deprecate them?

src/DataValues/DecimalValue.php
filbertkm: @thiemowmde I don't really care too much about the wording but https://github.com/DataValues/Number/pull/108 is what I had/have in mind. Using builders in DataValueSerializer is one alternative.

Hopefully we can remove this method entirely at some point soon

thiemowmde:
> […] why deprecate them?

Because the name is bad and new ones with better names should be introduced, if needed. Oh yea, and because I believe a deprecation phase is worth it. A rename would be a breaking change.

PR #108, opened 2017-06-26
Tweak comment in DecimalValue::newFromArray

filbertkm:
Using DataValueDeserializer is one alternative to this method.

### DIFF
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -355,10 +355,11 @@ class DecimalValue extends DataValueObject {
 	 *
 	 * @deprecated since 0.8.3. Static DataValue::newFromArray constructors like this are
 	 *  underspecified (not in the DataValue interface), and misleadingly named (should be named
-	 *  newFromArrayValue). Instead, use DataValue builder callbacks in @see DataValueDeserializer.
+	 *  newFromArrayValue). Instead, use the constructor or use DataValue builder
+	 *  callbacks. (e.g. @see DataValueDeserializer)
 	 *
 	 * @param mixed $data Warning! Even if this is expected to be a value as returned by
-	 *  @see getArrayValue, callers of this specific newFromArray implementation can not guarantee
+	 *  @see getArrayValue, callers of this specific newFromArray implementation cannot guarantee
 	 *  this. This is not guaranteed to be a string!
 	 *
 	 * @throws InvalidArgumentException if $data is not in the expected format. Subclasses of
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -356,7 +356,7 @@ class DecimalValue extends DataValueObject {
 	 * @deprecated since 0.8.3. Static DataValue::newFromArray constructors like this are
 	 *  underspecified (not in the DataValue interface), and misleadingly named (should be named
 	 *  newFromArrayValue). Instead, use the constructor or use DataValue builder
-	 *  callbacks. (e.g. @see DataValueDeserializer)
+	 *  callbacks (e.g. @see DataValueDeserializer).
 	 *
 	 * @param mixed $data Warning! Even if this is expected to be a value as returned by
 	 *  @see getArrayValue, callers of this specific newFromArray implementation cannot guarantee
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -355,8 +355,7 @@ class DecimalValue extends DataValueObject {
 	 *
 	 * @deprecated since 0.8.3. Static DataValue::newFromArray constructors like this are
 	 *  underspecified (not in the DataValue interface), and misleadingly named (should be named
-	 *  newFromArrayValue). Instead, use the constructor or use DataValue builder
-	 *  callbacks (e.g. @see DataValueDeserializer).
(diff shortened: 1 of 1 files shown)


manicki:
I like it! Stated that way it is less confusing (as using DataValueDeserializer is not given as the only option), which I was complaining about in https://github.com/DataValues/Number/issues/106.

src/DataValues/DecimalValue.php
```diff
@@ -355,10 +355,11 @@ public function getArrayValue() {
 	 *
 	 * @deprecated since 0.8.3. Static DataValue::newFromArray constructors like this are
 	 *  underspecified (not in the DataValue interface), and misleadingly named (should be named
-	 *  newFromArrayValue). Instead, use DataValue builder callbacks in @see DataValueDeserializer.
+	 *  newFromArrayValue). Instead, use the constructor or use DataValue builder
+	 *  callbacks. (e.g. @see DataValueDeserializer)
```
thiemowmde: I find the "or" problematic because the possibility to pass callbacks to DataValueDeserializer is not an alternative for having constructors. The callback must construct a DataValue object. Also I find the typography (brackets outside of the sentence) weird.

`Instead, use the default constructor or introduce a new static constructor with a better name, e.g. when needed for @see DataValueDeserializer.`

manicki: > Also I find the typography (brackets outside of the sentence) weird.

Oh, come on. Just move the full stop to the end:
> Instead, use the constructor or use DataValue builder callbacks (e.g. @see DataValueDeserializer).

What I find better in the version @filbertkm is that it is more generic. What you suggest instead is:
 - assuming one must use some static constructor in those deserializer callbacks. Maybe it is true in the end. But far too detailed for this particular place IMO.
 - IMO saying "introduce a new X but call it Y when you need it" is a bit odd thing to say in the deprecation notice about removing "X".

filbertkm: Use of DataValueDeserializers (builders) is one alternative to this deprecated method but this component can be used on it's own. The alternative w/o use of separate component is just construct the object directly

thiemowmde: > […] move the full stop to the end […]

Sure, that's what I'm asking for. Why do I get a "come on" for requesting this?

> assuming one must use some static constructor in those deserializer callbacks.

I don't get where I'm saying this. What I say is "Instead, use the default constructor **or** […]". There is an or. In most cases the default constructor will be enough.

> saying "introduce a new X […] in the deprecation notice about removing "X".

I don't get where I say this either. Again: a static constructor might be convenient, but it should not be called "newFromArray". Reasons for this are explained extensively in the posts above.

thiemowmde: @aude, I know that. Is this meant as a response to something @manicki or I said?

manicki:
Okay, added commit moving the period to the end of the sentence.
I believe we would find better use of our time than debating on punctuation or on what this or that way of phrasing things could be interpreted right.

Here is my review for this pull request:
 - the new message mentions the alternative of regular constructors. It might be stating the obvious but I think it is good to mention this. The existing comment does not do it, so this is good change IMO.
 - the new message expresses what the existing comment also does in a way that in my personal opinion is more clear and precise.
 - therefore I'm OK to merge this PR and adjust related messages.

I really do not intend to spend days on crafting the best depreciation comment message ever. But as the new comment was suggested, and it is in my opinion better than the one there is now, I think it is good to change it. My take on this PR would be that either:
 - there is a reason brought up why the suggested text is wrong (i.e. if it included information which is incorrect or misleading for users) - this is not a case as far as I can tell.
 - we accept the new comment message as suggested by @filbertkm , 
 - or there is a better alternative comment suggested which we like the best (the alternative suggested above by @thiemowmde is in my opinion not that good as what @filbertkm proposes). If there are other ideas, now it is a time to bring them up here!

thiemowmde:
I don't like to repeat myself, but it appears what I said was not heard: Stating "use the constructor or use DataValue builder callbacks" sounds like callbacks can act as replacements for constructors, which is not possible.

Please tell me what's wrong with `Instead, use the default constructor or introduce a new static constructor with a better name, e.g. when needed for @see DataValueDeserializer.`

manicki:
Okay, as you asked.

> Please tell me what's wrong with `Instead, use the default constructor or introduce a new static constructor with a better name, e.g. when needed for @see DataValueDeserializer.`

To me the second part of this sentence ("introduce a new static constructor with a better name, e.g. when needed") seems a really surprising thing to be said in the depreciation notice. Isn't what you're saying (in the second part, I know you mention the regular constructor in the first part) basically: "Don't use this. Instead introduce something new if you need. Just give it a good name"?

Reading it now, I would not have called __construct "default constructor" either. This is the only constructor exists in PHP, isn't it? But this is a minor detail.

thiemowmde:
"[Static constructor methods](https://google.com/search?q=static+constructor+methods)", short "[static constructors](https://google.com/search?q=static+constructors)". "[Default constructor](https://google.com/search?q=default+constructor)" seems to be much more ambiguous, but what else is `__construct`?

> Isn't what you're saying […] basically: […]

Yes, that's it. It appears you think there is something wrong with this, but I don#t get it.

mariushoch:
+1

JeroenDeDauw:
https://github.com/DataValues/Number/pull/125

PR #115, opened 2017-12-12
Stop outputting 100 decimal places for quantities

thiemowmde:
[Bug: T155910](https://phabricator.wikimedia.org/T155910)

### DIFF
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -91,22 +91,29 @@ class DecimalValue extends DataValueObject {
 			throw new InvalidArgumentException( '$number must not be NAN or INF.' );
 		}
 
-		if ( is_int( $number ) || $number === (float)(int)$number ) {
-			$decimal = strval( (int)abs( $number ) );
-		} else {
-			$decimal = trim( number_format( abs( $number ), 100, '.', '' ), '0' );
+		$decimal = strval( abs( $number ) );
+		$decimal = preg_replace_callback(
+			'/(\d*)\.(\d*)E([-+]\d+)/i',
+			function ( $matches ) {
+				list( , $before, $after, $exponent ) = $matches;
 
-			if ( $decimal[0] === '.' ) {
-				$decimal = '0' . $decimal;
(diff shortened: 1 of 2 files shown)


mariushoch:
Seems sensible, but will make the results depend on the [`precision` ini setting](https://secure.php.net/manual/en/ini.core.php#ini.precision). In WMF production this has the default value of 14.

I wonder whether you did this on purpose or whether we rather want to define the precision ourselves (which would mean using for example `printf('%0.14f', $theFloat)`, rather than `strval`).

mariushoch:
Note: Will probably also fix [T187888](https://phabricator.wikimedia.org/T187888).

thiemowmde:
Uh, thanks for bringing this up.

Note that PHPs `precision` setting is about **significant** digits. This is a concept that does not exist as a function in PHP (in contrast to JavaScript, which does have [`toPrecision`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision)). There are certainly ways to build such a function, but I wanted to avoid reimplementing something that already exists as a language feature. I intentionally rely on `strval` because it is the only way I can pull of PHPs build-in rounding to **significant** digits that fit what IEEE floating point numbers can represent.

For example, `printf( '%.30f', 1 / 3 )` outputs 0.333333333333333314829616256247, and `printf( '%.30f', 100000 / 3 )` outputs 33333.333333333335758652538061141968. In both cases, there is no fixed amount of meaningful digits after the decimal point, but (about) 16 significant digits that are meaningful.

PHPs `precision` setting is a little less, but still precise enough to describe the circumference of the earth (roughly 40,000 km) with a precision of 1 μm. That's quite literally a single piece of dust.

Note that reaching higher precisions is always possible by simply providing a string instead of a floating point number! All the code we are talking about here is exclusively about converting IEEE floats to strings.

mariushoch:
That sounds very well thought through, and definitely precise and robust enough for our purposes :)

Given this, I think it's safe to merge this!

PR #118, opened 2018-04-10
Release 0.10.0 with refined float to string conversion

thiemowmde:
I played around with the float-to-string conversion, tried a lot of different approaches I already tried before, and found a new one:
* I looked into PHP's "precision" setting. It defaults to 14. I found this problematic for two reasons:
  1. After #115 this code started to depend on local configuration.
  2. I found 14 not enough. ECMAScript for example outputs about 17 significant digits.
* I tried json_encode, but found it uses the same "precision" setting.
* I found a "serialize_precision" setting that defaults to 17. This fits much better to what this code is meant to do: The string is meant to be a serialization, not to be looked at by a human. With 17 places the output might end with 1 or 2 partially messed up digits. For example, you might end with "0.99999999999999989" instead of the expected "0.99999999999999999". This is because of the IEEE conversion. What's relevant: With 17 places the string is guaranteed to become the same IEEE floating point number when converted back.

This patch fixes two of the listed issues, and (I believe) avoids all others:
* The default of 17 significant digits is used and enforced.
* Because it now uses PHP's `serialize()`, the locale does not matter any more.

[Bug: T155910](https://phabricator.wikimedia.org/T155910)

### DIFF
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -91,7 +91,17 @@ class DecimalValue extends DataValueObject {
 			throw new InvalidArgumentException( '$number must not be NAN or INF.' );
 		}
 
-		$decimal = strval( abs( $number ) );
+		/**
+		 * As PHP has no other built-in way to format a float with a given number of significant
+		 * digits, use serialize(). While strval() would also work here, it is locale-dependent.
+		 * Intentionally enforce the bigger "serialize_precision" default of 17 to ensure a full
+		 * float-string-float roundtrip.
+		 * @see http://php.net/manual/en/ini.core.php#ini.serialize-precision
+		 */
+		$originalPrecision = ini_set( 'serialize_precision', '17' );
+		$decimal = substr( serialize( abs( $number ) ), 2, -1 );
+		ini_set( 'serialize_precision', $originalPrecision );
+
 		$decimal = preg_replace_callback(
 			'/(\d*)\.(\d*)E([-+]\d+)/i',
 			function ( $matches ) {
--- a/src/DataValues/DecimalValue.php
+++ b/src/DataValues/DecimalValue.php
@@ -92,38 +92,30 @@ class DecimalValue extends DataValueObject {
 		}
 
 		/**
-		 * As PHP has no other built-in way to format a float with a given number of significant
-		 * digits, use serialize(). While strval() would also work here, it is locale-dependent.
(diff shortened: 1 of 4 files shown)


src/DataValues/DecimalValue.php
```diff
@@ -91,7 +91,12 @@ private function convertToDecimal( $number ) {
 			throw new InvalidArgumentException( '$number must not be NAN or INF.' );
 		}
 
-		$decimal = strval( abs( $number ) );
+		// Enforce the default, see http://php.net/manual/en/ini.core.php#ini.serialize-precision
```
brightbyte: Would be nice to document why the strange detour via serialize() is needed.

thiemowmde: Uh, I'm not sure what would be helpful as a comment here. We could stick to strval, but would need to temporarily disable the locale then. json_encode is an other option. Do you think this is better?

mariushoch: Maybe:
```
// As PHP has no other built-in way to format a number with a given number of significant digits,
// use serialize. While strval would also work for this, it is locale-dependent.
```

mariushoch:
Looking at the PHP sources (https://github.com/php/php-src/blob/414aaa207cc59c279d3fd5ed83f6b92890e57023/Zend/zend_operators.c#L560), I think you could also use `sprintf("%.17G", $number);` here (although it is to note that this is probably not using the same code path as serialize does, which uses `gcvt` internally).

src/DataValues/DecimalValue.php
mariushoch: Maybe:
```
// As PHP has no other built-in way to format a number with a given number of significant digits,
// use serialize. While strval would also work for this, it is locale-dependent.
```

thiemowmde:
I updated the comment. Thanks!

sprintf can't be used as it does not allow to specify *significant* digits.

src/DataValues/DecimalValue.php
brightbyte: sprintf would probably be the cleaner solution, as proposed by marius below. If that's not possible, a comment should explain why.

thiemowmde:
I heavily updated the patch to use sprintf with %e now. That was new to me (and is not well documented, unfortunately). It indeed allows to format floats with a fixed number of **significant** digits, which is what we need here.

PR #121, opened 2018-10-29
DecimalMath::product: Round to net exceed 127 chars

mariushoch:
Bug: T208106

### DIFF
--- a/src/DataValues/DecimalMath.php
+++ b/src/DataValues/DecimalMath.php
@@ -61,6 +61,12 @@ class DecimalMath {
 		if ( $this->useBC ) {
 			$scale = strlen( $a->getFractionalPart() ) + strlen( $b->getFractionalPart() );
 			$product = bcmul( $a->getValue(), $b->getValue(), $scale );
+
+			$sign = $product[0] === '-' ? '' : '+';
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Not seeing anything wrong but am quite confused by the test, so can't +1

mariushoch:
> Not seeing anything wrong but am quite confused by the test, so can't +1

The test is esentially multiplying `0.3333…` (the closest we can get to that with our 127 chars) by `0.2222…` (the closest we can get to that with our 127 chars), and making sure that the resulting number exactly fits into the 127 char limit.

manicki:
Wouldn't some assert that strlen is (no more than) 127 be the even more clear description of the tested behaviour?

mariushoch:
> I don't get the `'+0.' . str_repeat( '2', 127 - 6 )` bit. That string is not 127 chars. Why the -6?

Actually this should be -4… two chars for "+" and ".", and two for the integer parts (the two "0").

I'll update this as suggested by @manicki.

JeroenDeDauw:
PR for releasing this: https://github.com/DataValues/Number/pull/120

PR #124, opened 2018-10-31
Test DecimalMath::product without bcmath

mariushoch:
And make sure the results are consistent with the bcmath output, by turning 1.0 * 5 into 5.0.

### DIFF
--- a/src/DataValues/DecimalMath.php
+++ b/src/DataValues/DecimalMath.php
@@ -98,7 +98,7 @@ class DecimalMath {
 			$product === floor( $product ) &&
 			$a->getFractionalPart() . $b->getFractionalPart() !== ''
 		) {
-			$product .= '.0';
+			$product = strval( $product ) . '.0';
 		}
 
 		return new DecimalValue( $product );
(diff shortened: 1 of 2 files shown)


src/DataValues/DecimalMath.php
```diff
@@ -59,16 +59,46 @@ public function getUseBC() {
...
+	private function productWithoutBC( DecimalValue $a, DecimalValue $b ) {
+		$product = $a->getValueFloat() * $b->getValueFloat();
+
+		// Append .0 for consistency, if the result is a whole number,
+		// but $a or $b were specified with decimal places.
+		if (
+			$product === floor( $product ) &&
+			$a->getFractionalPart() . $b->getFractionalPart() !== ''
+		) {
+			$product .= '.0';
```
JeroenDeDauw: Maybe good to do an explicit cast to string. I got a bit confused by the type changing implicitly here. Also suspect that when using PHP strict types this will cause some warning/error.

mariushoch: Good point… changed to an explicit string conversion.

JeroenDeDauw:
Looks good to me though I prefer if someone else does a sanity check before merging

PR #129, opened 2020-10-12
Update wikibase codesniffer version

chukarave:
for [T264874](https://phabricator.wikimedia.org/T264874)


micgro42:
I added a commit that should fix the license tag error. I hope you don't mind :)

Also, we should probably through the old versions in Travis away as done in a261249ab59ecf31658e863e25866577bb78d53e. I haven't added this commit here as I suspect that adding PHP7.3 and PHP7.4 might cause issues with PHPUnit, so for the sake for this PR, we maybe should just limit ourselves to dropping everything older than PHP7.2

lucaswerkmeister (approved):

PR #130, opened 2020-10-19
update phpunit to 8.5

rosalieper:
- Use phpunit TestCase
- Fix failing test due to phpunit upgrade.
- Add new testBase abstract classes
- Change`getAmount()->getValue` to `getAmount()->getValueFloat` in tests. In this case
comparing float values seem more reasonable than comparing stringyfied decimals.
- See: https://phabricator.wikimedia.org/T260915


rosalieper:
fails due to dependency on outdated DataValue/DataValue

tests/DataValues/QuantityValueTest.php
```diff
@@ -138,10 +139,14 @@ public function newFromNumberProvider() {
...
+		$expectedValue = new DecimalValue( $value->getValueFloat() );
+		$expectedValue1 = new DecimalValue( $value1->getValueFloat() );
+		$expectedValue2 = new DecimalValue( $value2->getValueFloat() );
+
 		$unit = 'm';
 		yield [
 			$value->getValueFloat(), $unit, $value1->getValueFloat(), $value2->getValueFloat(),
-			new QuantityValue( $value, $unit, $value1, $value2 )
+			new QuantityValue( $expectedValue, $unit, $expectedValue1, $expectedValue2 )
 		];
```
itamargiv, line 150: I'm not sure we should change the test content like that, what's happening now (afaict) is that we are passing the expected result into the constructor, by essentially assuming the result we expect to get. What issue were you trying to resolve by this?

tests/DataValues/DataValuesTest.php
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Base for unit tests for DataValue implementing classes.
+ *
+ * @since 0.1
+ *
+ * @license GPL-2.0-or-later
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+abstract class DataValuesTest extends TestCase {
```
itamargiv, line 22: I'm not convoinced we really need to copy all these tests to here, I wonder what @lucaswerkmeister thinks about it...

itamargiv (changes requested):
Yay \o/ The tests finally pass! I do have some concerns regarding the changes we made to some of the tests there, but apart from that, great work, thank you!

tests/DataValues/QuantityValueTest.php
rosalieper: I think the variable name "$expectedValue is not a good one" actually the test had a small bug and for some reason was passing on phpunit version 4 undetected. the actual computation changes a decimal to a float then converts the float output to decimal again. but the expected computation change to decimal and return the decimal. Those 2 computations give different values when the input is a negative decimal. So I change the expected computation to be similar to the actual by adding a conversion to float before converting to Decimal. I believe the variable name is confusing here. Maybe this long comment is also confusing. :(

We can jump in a call and I show you what I mean if not clear.

tests/DataValues/DataValuesTest.php
lucaswerkmeister: Yeah, I don’t think copying the tests is the best way forward – after all, the DataValues release notes say:

> * Removed `DataValueTest` (create a copy if you need it, though better refactor away the bad design)

Some of the abstract tests here can just be removed as completely pointless, for instance:

```php
$this->assertInstanceOf( $this->getClass(), $dataItem );

$this->assertInstanceOf( Immutable::class, $value );

$this->assertTrue( true );
```

It’s better to put specific, meaningful tests in the particular test classes.

tests/DataValues/QuantityValueTest.php
lucaswerkmeister: I still don’t understand what is happening here, or what the difference between variables like `$value` and `$decimalValue` is supposed to be when they’re both instances of `DecimalValue` :/

toban:
LGTM. 

One difference I noticed running the tests locally on master and then this branch is there is now a Warning about the following.

```
1) Warning
The data provider specified for ValueFormatters\Test\QuantityHtmlFormatterTest::testGivenHtmlCharacters_formatEscapesHtmlCharacters is invalid.
PHPUnit\Framework\MockObject\RuntimeException: Trying to configure method "getUncertaintyMargin" which cannot be configured because it does not exist, has not been specified, is final, or is static
/home/toan/Documents/dev/Number/tests/ValueFormatters/QuantityHtmlFormatterTest.php:75
/home/toan/Documents/dev/Number/tests/ValueFormatters/QuantityHtmlFormatterTest.php:171

```

rosalieper:
> LGTM.
> 
> One difference I noticed running the tests locally on master and then this branch is there is now a Warning about the following.
> 
> ```
> 1) Warning
> The data provider specified for ValueFormatters\Test\QuantityHtmlFormatterTest::testGivenHtmlCharacters_formatEscapesHtmlCharacters is invalid.
> PHPUnit\Framework\MockObject\RuntimeException: Trying to configure method "getUncertaintyMargin" which cannot be configured because it does not exist, has not been specified, is final, or is static
> /home/toan/Documents/dev/Number/tests/ValueFormatters/QuantityHtmlFormatterTest.php:75
> /home/toan/Documents/dev/Number/tests/ValueFormatters/QuantityHtmlFormatterTest.php:171
> ```

warning fixed.

tests/DataValues/QuantityValueTest.php
rosalieper: I changed the variable name to `decimalValueFromFloat`. Is that better?

itamargiv (approved):
:100: Looking good! thanks for the effort :+1:

lucaswerkmeister (approved):
Looks good to me! I’m not sure how we want to merge this – I think it can all be squashed into one commit?

PR #137, opened 2022-10-14
Implement _serialize()/__unserialize()

lucaswerkmeister:
Bug: T301249


lucaswerkmeister:
The float-related test failures are due to the PHPUnit version – I can reproduce them locally starting with [8.5.24](https://github.com/sebastianbergmann/phpunit/blob/main/ChangeLog-8.5.md#8524---2022-03-05---standwithukraine), but 8.5.23 is fine.

lucaswerkmeister:
Rebased onto #138, which should fix the CI issues.

PR #139, opened 2022-10-21
Remove DATAVALUES_NUMBER_VERSION

lucaswerkmeister:
We’d already forgotten to update it with the previous two releases, and we’ve also removed such constants from other data-values libraries.


micgro42 (approved):
Yeah, dependency management should happen via composer. Not by checking some version constant in code

PR #141, opened 2022-10-21
Relax data-values/data-values version constraint

lucaswerkmeister:
~3.0.0 means that 3.1.0 is excluded, which we don’t want; use ~3.0 instead for consistency with the other versions in this constraint. (Though I’d personally prefer to just use the caret operator.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,10 @@ the [Wikidata project](https://www.wikidata.org/).
 
 ## Release notes
 
+### 0.12.1 (2022-10-21)
+
+* Allow use with data-values/data-values 3.1.0
+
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
I would tag the merge commit of this as 0.12.1 directly, with out a separate “release” commit, if that’s okay with everyone.

micgro42 (approved):
I wonder if it would make sense to switch to use `^x.y.z` when we want to allow minor-level updates and only use `~x.y.z` when we want to limit updates to the patch level. That way it would be compatible with npm.

But that would be a discussion on a meta level. And probably for a different team.

lucaswerkmeister:
Tagged.

PR #142, opened 2022-10-21
Fix hashes after serialization change

lucaswerkmeister:
Because the serialization of quantity values involves complex objects (`DecimalValue` instances), the `getSerializationForHash()` implementation we inherit from `DataValueObject` isn’t good enough: we need to override it to also call `getSerializationForHash()` on the inner `DecimalValue` instances (amount, upper bound, lower bound).

The hashes for the tests were obtained inside a php:7.2 Docker container, after installing the specified versions with composer, inside a `php -a` shell.

Bug: T301249

### DIFF
--- a/src/DataValues/QuantityValue.php
+++ b/src/DataValues/QuantityValue.php
@@ -130,6 +130,26 @@ class QuantityValue extends UnboundedQuantityValue {
 		$this->__construct( $amount, $unit, $upperBound, $lowerBound );
 	}
 
+	public function getSerializationForHash(): string {
+		// mimic a legacy serialization of __serialize() (amount + unit + upperBound + lowerBound)
+		$amountSerialization = method_exists( $this->amount, 'getSerializationForHash' )
+			? $this->amount->getSerializationForHash()
+			: serialize( $this->amount );
+		$unitSerialization = serialize( $this->unit );
+		$upperBoundSerialization = method_exists( $this->upperBound, 'getSerializationForHash' )
+			? $this->upperBound->getSerializationForHash()
+			: serialize( $this->upperBound );
+		$lowerBoundSerialization = method_exists( $this->lowerBound, 'getSerializationForHash' )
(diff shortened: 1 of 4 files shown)


tests/DataValues/QuantityValueTest.php
```diff
@@ -458,4 +458,26 @@ public function transformProvider() {
...
+		$this->assertSame( $hash, $quantity->getHash() );
+	}
+
+	public function instanceWithHashProvider(): iterable {
+		// all hashes obtained from data-values/data-values==3.0.0 data-values/number==0.11.1 under PHP 7.2.34
+		yield '10+-1' => [
+			QuantityValue::newFromNumber( '+10', '1', '+11', '+9' ),
+			'2eb812346e6bbc1bc47ed7da130bfa5d',
+		];
+		yield '500 miles, or a little bit more' => [
```
lucaswerkmeister, line 472: reference: https://www.ibiblio.org/harris/500milemail.html :)

src/DataValues/QuantityValue.php
```diff
@@ -130,6 +130,26 @@ public function __unserialize( array $data ): void {
 		$this->__construct( $amount, $unit, $upperBound, $lowerBound );
 	}
 
+	public function getSerializationForHash(): string {
+		// mimic a legacy serialization of __serialize() (amount + unit + upperBound + lowerBound)
+		$amountSerialization = method_exists( $this->amount, 'getSerializationForHash' )
```
lucaswerkmeister, line 135: Note: `DecimalValue` is itself a `DataValueObject` subclass, so it inherits `getSerializationForHash()` from there. The inherited version is fine, so we don’t need to touch `DecimalValue.php`.

(And the `method_exists()` check should in theory ensure that this still works with pre-3.1.0 DataValues versions, but I haven’t tested that.)

micgro42 (approved):

PR #145, opened 2024-12-06
Support mediawiki-codesniffer@45.0.0, PHP 8.4 readiness

codders:
This repository is currently using version 34 of the Mediawiki coding style phpcs rules. It should be upgraded to the latest version.

Besides being simply different, the newer versions of the rules introduce, for example,
MediaWiki.Usage.NullableType.ExplicitNullableTypes, which enforces compliance with coming deprecation rules in PHP 8.4 concerning nullable types. If the code does not remove the soon-to-be-deprecated form of nullable type declarations, the library will no longer be usable in Mediawiki for PHP 8.4 deployments.

Bug: T379481

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: php
 dist: bionic
 
 php:
-  - 7.2
-  - 7.3
   - 7.4
   - 8.0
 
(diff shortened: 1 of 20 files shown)


yerdua (approved):
lgtm!

PR #146, opened 2024-12-09
Bump version to 0.13.0

codders:
* Drop support for PHP 7.2, 7.3
 * Upgrade codesniffer rules to current `mediawiki/mediawiki-codesniffer` version (45.0.0)
 * Make nullable type parameter declarations explicit for compatibility with PHP 8.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,12 @@ the [Wikidata project](https://www.wikidata.org/).
 
 ## Release notes
 
+### 0.13.0 (2024-12-10)
+
+* Drop support for PHP 7.2, 7.3
+* Upgrade codesniffer rules to current `mediawiki/mediawiki-codesniffer` version (45.0.0)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,6 +49,12 @@ the [Wikidata project](https://www.wikidata.org/).
 
 ## Release notes
 
+### 0.13.0 (2024-12-10)
```
lucaswerkmeister, line 52: Is this intentionally set to tomorrow’s date? ^^

codders: yes - I haven't had so many same-day reviews :) I figure if the review gets through I can hit the 'create release' button tomorrow.

lucaswerkmeister: LGTM, but IMHO we shouldn’t merge it today then.

lucaswerkmeister (approved):

